PR #1, opened 2016-09-17
use latest versions of google closure from npm

Hamms:
also remove local versions of closure tools and closure compiler.

The penultimate commit removes the entire `closure-compiler-read-only` directory and is therefore huge. I recommend focusing your review on only [the first two](https://github.com/code-dot-org/blockly/compare/staging...e432039f24e56c2b87b823f0ef074418aab30a49) and [the final](https://github.com/code-dot-org/blockly/pull/1/commits/7e72ca24dbd9b1acaa328394e239f9af67586d3e) commits


Hamms:
Also, for posterity; this change and enabling linting are the two things I really wanted to get done before bumping this up to v1.0

package.json
```diff
@@ -18,8 +18,9 @@
   "scripts": {
     "build": "./deploy.sh",
     "build:dev": "./deploy.sh debug",
+    "preversion": "npm run test",
+    "version": "npm run build:dev && npm run build && git add -A build-output",
```
islemaster: Does this replace the default `npm version` behavior?  If not, does this run before or after it?

Hamms: it runs after bumping the version tag in `package.json` but before committing and tagging

Hamms: https://docs.npmjs.com/cli/version has the exact order of execution

islemaster: Sweet - LGTM!

bcjordan:
Fantastic! LGTM

PR #3, opened 2016-09-26
Remove "use strict" from blockly_compressed.js

Hamms:
Its presense for some reason breaks apps tests


Hamms:
This PR is necessary for https://github.com/code-dot-org/code-dot-org/pull/10794 to pass unit tests

PR #7, opened 2016-10-25
Fix extra whitespace in webkit

Hamms:
A little backstory: webkit browsers have a [long-existing bug](https://bugs.chromium.org/p/chromium/issues/detail?id=230599) where
shapes with control points take those points into account when
calculating the BBox. We had some workarounds in place to detect and
handle this bug, but Chrome [recently fixed this
issue](https://chromium.googlesource.com/chromium/src.git/+/8f4c8e6d2f1260d68f387b37f078457e4153bbb4) _only for paths_. This means that a container with an offending path
will still return the bugged BBox.

Unfortunately, our test was checking just a path, while our actual code
was using a container with a path, resulting in our test failling to
detect the presence of the bug.

The simple solution is to update our test to more accurately reflect our
real-life use case.

### DIFF
--- a/core/ui/block_space/block_space_editor.js
+++ b/core/ui/block_space/block_space_editor.js
@@ -530,6 +530,7 @@ Blockly.BlockSpaceEditor.prototype.detectBrokenControlPoints = function() {
 
      NOTE:
      chromium fixed this bug for path elements in commit 8f4c8e;
+     (https://chromium.googlesource.com/chromium/src.git/+/8f4c8e6d2f1260d68f387b37f078457e4153bbb4)
      unfortunately, the bug still presents itself when getBBox is called
      on containers that CONTAIN such a path element, which is our actual
      use case. Thus, we draw the shape within a container and check the
--- a/core/ui/block_space/block_space_editor.js
+++ b/core/ui/block_space/block_space_editor.js
@@ -527,27 +527,35 @@ Blockly.BlockSpaceEditor.prototype.detectBrokenControlPoints = function() {
      shape that is 50px high, and has a control point that sticks up by 5px.
      If the getBBox function returns a height of 55px instead of 50px, then
      this browser has broken control points.
+
+     NOTE:
+     chromium fixed this bug for path elements in commit 8f4c8e;
+     unfortunately, the bug still presents itself when getBBox is called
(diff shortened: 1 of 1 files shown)


Hamms:
FYI @breville and @davidsbailey

breville:
nice fix!

core/ui/block_space/block_space_editor.js
```diff
@@ -527,27 +527,35 @@ Blockly.BlockSpaceEditor.prototype.detectBrokenControlPoints = function() {
      shape that is 50px high, and has a control point that sticks up by 5px.
      If the getBBox function returns a height of 55px instead of 50px, then
      this browser has broken control points.
+
+     NOTE:
+     chromium fixed this bug for path elements in commit 8f4c8e;
+     unfortunately, the bug still presents itself when getBBox is called
+     on containers that CONTAIN such a path element, which is our actual
```
bcjordan: Awesome workaround :). Reminds me of the [Chrome 34 SVG clippath fix](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/code-studio/initApp/chrome34Fix.js#L10) (wrap in outer svg!)

core/ui/block_space/block_space_editor.js
```diff
@@ -527,27 +527,35 @@ Blockly.BlockSpaceEditor.prototype.detectBrokenControlPoints = function() {
      shape that is 50px high, and has a control point that sticks up by 5px.
      If the getBBox function returns a height of 55px instead of 50px, then
      this browser has broken control points.
+
+     NOTE:
+     chromium fixed this bug for path elements in commit 8f4c8e;
```
bcjordan: Googling "8f4c8e chromium" didn't come up with anything—is there a link to a web view of that commit that could be included? NBD if not.

bcjordan:
(Shouldn't block merging—) is there a good place to report chromium's change's incompleteness? Was the initial fix associated with an issue?

Hamms:
Reporting the issue definitely seems worthwhile; I'm not certain of the best place to do so, but I'm sure I can find something.

bcjordan:
Looks like https://bugs.chromium.org/p/chromium/issues/detail?id=230599

PR #8, opened 2016-11-10
Animate Angle Helper

Hamms:
When the angle helper's angle is changed by an external field change,
smoothly animate the transition rather than jumping directly to the new
value.

### Before

before

### After

after


bcjordan:
Love it! LGTM!

PR #9, opened 2016-11-10
Highlight Angler Helper's "primary" half

Hamms:
Rather than treating all 360 degrees of the angle helper as being equal,
subtly nudge the students toward making turns in the 0-180 region by
reducing the opacity of the tickmarks on the other half.

image
image


bcjordan:
Awesome UX improvement, LGTM!

PR #10, opened 2016-11-10
Fix Angle Helper arc "jumping"

Hamms:
Previously, the Angle Helper arc would jump around erratically when
changing the angle. Turns out that rounding the values used to generate
the arc caused some erratic behaviour.

before

Rounding to two decimal places instead of to int results in a much smoother look, while still giving us some testable amount of determinism.

after


bcjordan:
Yay, LGTM!

PR #11, opened 2016-11-11
Fix block bumping on initialization

Hamms:
In a situation in which a top-level block with two input statements has
two or more blocks in the first one and a block containing a dropdown in
the second one, the blocks in the first statement end up getting bumped
out of position.

image

This is because the dropdown, upon initialization, checks for nearby
connections and bumps them away. And because the parent block is not yet
initialized when the dropdown block is being initialized, it sees the
blocks in the other statement as being positioned right on top of them,
and so bumps them away.

image

Blocks with parents avoid this by being rendered before their children
get rendered, and top-level blocks with previous connections (ie, blocks
without parents that really should have parents) avoid this by making a
call to `setIsUnused` before initializing their children, which has the
side effect of making sure that they get rendered before their children
do.

image

The simple solution is to make sure that _all_ blocks get rendered
before their children do, by adding a call to `this.render()` to the end
of `Block.initSvg`.

image

### DIFF
--- a/core/ui/block.js
+++ b/core/ui/block.js
@@ -266,6 +266,7 @@ Blockly.Block.prototype.initSvg = function() {
   this.setCurrentlyHidden(this.currentlyHidden_);
   this.moveToFrontOfMainCanvas_();
   this.setIsUnused();
+  this.render();
 };
 
 /**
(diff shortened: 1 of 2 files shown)


bcjordan:
LGTM, nice catch!

PR #12, opened 2016-11-15
Add support for inputCount attribute

balderdash:


core/utils/xml.js
```diff
@@ -369,6 +372,10 @@ Blockly.Xml.domToBlock = function(blockSpace, xmlBlock) {
   if (userCreated) {
     block.userCreated = (userCreated === 'true');
   }
+  var inputCount = xmlBlock.getAttribute('inputCount');
+  if (inputCount) {
+    block.setInputCount(inputCount);
```
Hamms: I'd very much prefer this to be something defined on Blockly.Block, rather than a method implemented by a single type of block. Otherwise, we could end up getting some difficult-to-interpret errors if this attribute were set on a different kind of block.

balderdash: Oops, deleted my reply here.

I added a generic implemention, but kept the subclass one since I need that too.

balderdash:
Added a context menu item by adding support for a "getCustomContextMenuItems" method defined in the subclass.

PR #13, opened 2016-11-30
Remove highlight when removing input

balderdash:


balderdash:
It wasn't working before, but I couldn't find a place where we removed inputs when it could have been highlighted, e.g. reducing the number of parameters a function took never happened when you were in the process of connecting a block to a input of the function call block.

I noticed this issue only when I started working on the resizing join block.

PR #14, opened 2016-11-30
Add pending connection/disconnection hook

balderdash:


core/ui/block.js
```diff
@@ -1232,6 +1239,9 @@ Blockly.Block.prototype.moveBlockBeingDragged_ = function (mouseX, mouseY) {
   }
 };
 
+Blockly.Block.prototype.pendingConnection =
```
Hamms: could we get a block comment here explaining that this is intended to be overridden?

balderdash:
The join block is supposed to expand when you drag another block near the bottom of an existing block. Dragging it close enough to the last existing input to trigger a highlight is a pretty close approximation of that, so I'm using that instead of adding a general mousemove handler.

I like that the join block looks like it's exactly the right size most of the time, and only expands when you might need it to.

core/ui/block.js
balderdash: Done.

PR #15, opened 2016-12-01
Optionally call changehandler in setTitleValue

balderdash:
FieldDropdown's `changeHandler` is currently only called when the dropdown's value is changed through UI interaction. I'm adding an option to call it when it's changed by a call to `setTitleValue` as well.

I'll be using this option to fix the "Text in “When Any <X>” block doesn’t get properly updated on refresh" bug.

### DIFF
--- a/core/ui/block.js
+++ b/core/ui/block.js
@@ -1844,6 +1844,12 @@ Blockly.Block.prototype.getTitleValue = function(name) {
  */
 Blockly.Block.prototype.setTitleValue = function(newValue, name) {
   var title = this.getTitle_(name);
+  if (title.changeHandler_) {
+    var override = title.changeHandler_(newValue);
+    if (override) {
+      newValue = override;
+    }
+  }
   if (title) {
     title.setValue(newValue);
   } else {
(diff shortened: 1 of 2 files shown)


core/ui/block.js
```diff
@@ -1844,6 +1844,12 @@ Blockly.Block.prototype.getTitleValue = function(name) {
  */
 Blockly.Block.prototype.setTitleValue = function(newValue, name) {
   var title = this.getTitle_(name);
+  if (title.alwaysCallChangeHandler && title.changeHandler_) {
+    var override = title.changeHandler_(newValue);
+    if (override) {
+      newValue = override;
+    }
+  }
```
Hamms: I'm not sure I love this logic living here, and I definitely don't love it referencing the private value `changeHandler_` on title.

Could this go in `title.setValue`? Is that used elsewhere?

balderdash: Moved. I just had to add a bit of logic to the dropdown callback to avoid calling the handler twice, since that callback uses `setValue`.

PR #18, opened 2016-12-06
Reposition fields on scroll

Hamms:
Most (all?) of our editable fields in blockly don't reposition
themselves when the blockspace scrolls. (You can see an example of this [here](https://studio.code.org/s/grade1/stage/12/puzzle/9)) This solution adds a scroll
event which they can listen to to update themselves.

TODO: apply this approach to all fields, not just the rectangular
dropdown.

### DIFF
--- a/core/ui/fields/field_rectangular_dropdown.js
+++ b/core/ui/fields/field_rectangular_dropdown.js
@@ -158,7 +158,7 @@ Blockly.FieldRectangularDropdown.prototype.setArrowDirection_ = function(up) {
 };
 
 Blockly.FieldRectangularDropdown.prototype.showMenu_ = function() {
-  Blockly.WidgetDiv.show(this, this.generateMenuClosedHandler_());
+  this.showWidgetDiv_();
   this.menu_ = this.createMenuWithChoices_(this.choices_);
   goog.events.listen(this.menu_, goog.ui.Component.EventType.ACTION, this.generateMenuItemSelectedHandler_());
   this.addPositionAndShowMenu(this.menu_);
@@ -166,10 +166,6 @@ Blockly.FieldRectangularDropdown.prototype.showMenu_ = function() {
 };
 
 Blockly.FieldRectangularDropdown.prototype.hideMenu_ = function() {
-  if (this.blockSpaceScrolledListenKey_) {
-    this.sourceBlock_.blockSpace.events.unlistenByKey(this.blockSpaceScrolledListenKey_);
-    this.blockSpaceScrolledListenKey_ = null;
-  }
 
   this.pointArrowDown_();
 };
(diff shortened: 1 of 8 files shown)


core/ui/fields/field_rectangular_dropdown.js
```diff
@@ -251,11 +258,33 @@ Blockly.FieldRectangularDropdown.prototype.addPositionAndShowMenu = function (me
...
-  Blockly.WidgetDiv.position(menuPosition.x, menuPosition.y, windowSize, scrollOffset);
+  var positionBelow = multipleColumns;
+  this.updatePosition(positionBelow);
+
+  // if we are attached to a block, recalculate our position when the
+  // block's blockspace gets scrolled.
+  if (this.sourceBlock_) {
+    var events = this.sourceBlock_.blockSpace.events;
+
+    if (!this.blockSpaceScrolledListenKey_) {
```
bcjordan: Nit: can maybe fold into prior conditional

Hamms:
My one hesitation against putting the logic into WidgetDiv is that this logic requires there to be a BlockSpace associated with the calling widget, and that's not assumption I'd like to bake into WidgetDiv.

I think a better place to try and fit this logic in would be in the definition of the master `Field` class from which all of these implementations inherit.

bcjordan:
Yes, that sounds better!

Hamms:
@bcjordan PTAL

tests/fields_test.js
```diff
@@ -0,0 +1,78 @@
...
+      '<xml>' +
+      '  <block type="math_number">' +
+      '    <title name="NUM">0</title>' +
+      '  </block>' +
+      '</xml>'));
+
+  var block = blockSpace.getTopBlocks()[0];
+  var field = block.getTitle_("NUM");
+
+  window.positionWidgetDivCalled = false;
```
bcjordan: Can just be `var positionWidgetDivCalled`?

core/ui/fields/field_rectangular_dropdown.js
```diff
@@ -158,13 +158,18 @@ Blockly.FieldRectangularDropdown.prototype.setArrowDirection_ = function(up) {
...
-  Blockly.WidgetDiv.show(this, this.generateMenuClosedHandler_());
+  this.showWidgetDiv_();
   this.menu_ = this.createMenuWithChoices_(this.choices_);
   goog.events.listen(this.menu_, goog.ui.Component.EventType.ACTION, this.generateMenuItemSelectedHandler_());
   this.addPositionAndShowMenu(this.menu_);
   this.pointArrowUp_();
 };
 
+Blockly.FieldRectangularDropdown.prototype.hideMenu_ = function() {
+
```
bcjordan: Nit: bonus `\n`

tests/fields_test.js
Hamms: Oh, it can! This was an artifact of iteration on the best way to override the prototype method.

PR #19, opened 2016-12-09
Stop running bumpNeighbours for blocks being dragged in from the toolbox

balderdash:
Right now, dragging certain blocks in from the toolbox causes existing blocks at the top left of the workspace to get bumped out of position. This happens the moment you mousedown on the toolbox block, not when you place it somewhere (which is when neighbor bumping is supposed to happen).

This happens because `bumpNeighbours_` is called indirectly by `domToBlock` while the newly created block is still sitting at `0,0`. `bumpNeighbors_` normally terminates immediately during block dragging, but the `onMouseDown` handler doesn't have a chance to run until the block is initialized, so I'm making `domToBlock` temporarily disable block bumping when needed.

### DIFF
--- a/core/ui/block.js
+++ b/core/ui/block.js
@@ -81,6 +81,7 @@ Blockly.Block = function(blockSpace, prototypeName, htmlId) {
   this.nextConnectionDisabled_ = false;
   this.collapsed_ = false;
   this.dragging_ = false;
+  this.bumpingEnabled_ = true;
 
   // Used to hide function blocks when not in modal workspace. This property
   // is not serialized/deserialized.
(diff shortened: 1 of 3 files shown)


core/ui/block.js
```diff
@@ -1290,13 +1291,21 @@ Blockly.Block.prototype.generateReconnector_ = function(earlierConnection) {
...
+};
+
 /**
  * Bump unconnected blocks out of alignment.  Two blocks which aren't actually
  * connected should not coincidentally line up on screen.
  * @private
  */
 Blockly.Block.prototype.bumpNeighbours_ = function() {
-  if (Blockly.Block.isDragging() || !Blockly.BUMP_UNCONNECTED) {
+  if (Blockly.Block.isDragging() || !Blockly.BUMP_UNCONNECTED || !this.bumpingEnabled_) {
```
Hamms: if the problem is that the block is being dragged upon creation, isn't the real problem that `Blockly.Block.isDragging()` isn't false at this point?

balderdash:
Closing in favor of https://github.com/code-dot-org/blockly/pull/20

PR #27, opened 2017-02-09
Add 'onMainBlockSpaceCreated' helper

Hamms:
Rather than expecting other components to listen on the
MAIN_BLOCK_SPACE_CREATED event and handle all the associated logic
themselves, we do it internally.

### DIFF
--- a/core/ui/block_space/block_space.js
+++ b/core/ui/block_space/block_space.js
@@ -115,9 +115,8 @@ Blockly.BlockSpace.EVENTS = {};
 
 /**
  * Called after the mainBlockSpace has been initialized and assigned to
- * the Blockly.mainBlockSpace global attribute; can be used by
- * components that want to make their own one-off blockspaces to ensure
- * that they're preserving initialization order.
+ * the Blockly.mainBlockSpace global attribute, used by
+ * onMainBlockSpaceCreated
  * @type {string}
  */
 Blockly.BlockSpace.EVENTS.MAIN_BLOCK_SPACE_CREATED = 'mainBlockSpaceCreated';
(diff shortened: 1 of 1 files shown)


core/ui/block_space/block_space.js
```diff
@@ -217,6 +216,22 @@ Blockly.BlockSpace.createReadOnlyBlockSpace = function (container, xml, opt_opti
...
+ * Simple helper for situations where we depend on the mainBlockSpace
+ * having already been created; most useful for delaying components that
+ * want to make their own one-off blockspaces to ensure that they're
+ * preserving initialization order.
+ */
+Blockly.BlockSpace.onMainBlockSpaceCreated = function (callback) {
+  if (Blockly.mainBlockSpace) {
+    callback();
+  } else {
+    document.addEventListener(Blockly.BlockSpace.EVENTS.MAIN_BLOCK_SPACE_CREATED, function () {
```
joshlory: Nit: can pass in `callback` directly here.  (Yes I messed this up in my other comment 😛).

core/ui/block_space/block_space.js
```diff
@@ -217,6 +216,22 @@ Blockly.BlockSpace.createReadOnlyBlockSpace = function (container, xml, opt_opti
 };
 
 /**
+ * Simple helper for situations where we depend on the mainBlockSpace
+ * having already been created; most useful for delaying components that
+ * want to make their own one-off blockspaces to ensure that they're
+ * preserving initialization order.
```
joshlory: Nit: worth mentioning that, if the mainBlockSpace already exists the callback will be called immediately?

PR #29, opened 2017-02-10
Add option to disable editing of the number of branches in an if/else block

joshlory:
Following the pattern of `Blockly.disableParamEditing`.


joshlory:
`Blockly.disableIfElseEditing = false` (the default)

screen shot 2017-02-10 at 4 13 13 pm

`Blockly.disableIfElseEditing = true`

screen shot 2017-02-10 at 4 13 39 pm

PR #35, opened 2017-03-17
Move blocks to a top-level SVG while dragging

joshlory:
Allow blocks to be dragged outside the extents of their block space SVG.  Also ensures that the block is always at the top of the z-index while dragging.

### DIFF
--- a/core/ui/block_space/block_space.js
+++ b/core/ui/block_space/block_space.js
@@ -337,7 +337,7 @@ Blockly.BlockSpace.prototype.createDom = function() {
       id: 'blocklyDragCanvas',
       width: '100%',
       height: '100%',
-      style: 'pointer-events: none; position: absolute; top: 0; left: 0;'
+      style: 'pointer-events: none; position: absolute; top: 0; left: 0; z-index: 10;'
     }, document.body);
   }
   this.svgDragCanvas_ = Blockly.createSvgElement('g', {'class': 'svgDragCanvas'}, Blockly.dragSvg);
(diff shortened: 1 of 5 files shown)


core/ui/block_space/block_space.js
```diff
@@ -328,9 +328,22 @@ Blockly.BlockSpace.prototype.createDom = function() {
...
+    Blockly.dragSvg = Blockly.createSvgElement('svg', {
+      id: 'blocklyDragCanvas',
+      width: '100%',
+      height: '100%',
+      style: 'pointer-events: none; position: absolute; top: 0; left: 0;'
+    }, document.body);
+
+    Blockly.dragCanvas = Blockly.createSvgElement('g', {'class': 'svgDragCanvas'}, Blockly.dragSvg);
+  }
+  this.svgDragCanvas_ = Blockly.dragCanvas;
```
Hamms: damn, I love that this is so straightforward!

Hamms: Has this been manually tested with the function editor to make sure that two blockspaces using the same drag canvas is fine?

joshlory: It works correctly with the main block space / function editor block space, but I need to fix a few bugs with flyouts and mutators 😕.

Hamms: blocking bugs?

joshlory: Yeah, working to fix them today.

joshlory:
I've given each drag canvas a separate `<g>` element to more closely match the setup before moving the `<g>` elements to a separate SVG.  This should take care of a few lingering issues where the offset of the drag canvas is stale when dragging in a mutator, immediately followed by dragging in the main block space.  PTAL!

PR #36, opened 2017-03-27
i18n sync 2017-03-27

Hamms:


tanyaparker:
This doesn't need to include a blockly version bump?

Hamms:
The version will get bumped after merge

PR #37, opened 2017-04-04
Build source maps for blockly_compressed.js

joshlory:
Allows easier debugging on New Relic.  Without source maps:

source-maps-before

After adding the source map:

source-maps-after

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   },
   "devDependencies": {
     "phantomjs-prebuilt": "^2.1.3",
-    "google-closure-compiler": "^20160911.0.0",
+    "google-closure-compiler": "^20170218.0.0",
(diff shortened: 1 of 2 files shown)


joshlory:
I haven't done any integration with New Relic's API yet — for now I'm manually dragging and dropping the source map into New Relic.  It's not the best user experience but they make it fairly seamless.

package.json
```diff
@@ -25,7 +25,7 @@
   },
   "devDependencies": {
     "phantomjs-prebuilt": "^2.1.3",
-    "google-closure-compiler": "^20160911.0.0",
+    "google-closure-compiler": "^20170218.0.0",
```
joshlory: @Hamms any concerns with updating the version of the Closure Compiler?

PR #41, opened 2017-04-11
Unselecting a block should always terminate drag

joshlory:
Fixes multiple touch points bug on iOS while dragging a block.

Repro on iOS:
 - Start dragging a block.
 - Use a second finger to start panning the page.
 - Release when neither finger is touching the dragged block.

https://rpm.newrelic.com/accounts/501463/browser/3787364/js_errors#id=497715591

### DIFF
--- a/core/ui/block.js
+++ b/core/ui/block.js
@@ -399,6 +399,7 @@ Blockly.Block.prototype.unselect = function() {
   if (!this.svg_) {
     throw 'Block is not rendered.';
   }
+  Blockly.BlockSpaceEditor.terminateDrag_();
   Blockly.selected = null;
   this.svg_.removeSelect();
   this.svg_.removeSpotlight();


core/ui/block.js
```diff
@@ -399,6 +399,7 @@ Blockly.Block.prototype.unselect = function() {
   if (!this.svg_) {
     throw 'Block is not rendered.';
   }
+  Blockly.BlockSpaceEditor.terminateDrag_();
```
Hamms: why call the BlockSpaceEditor method and not `Blockly.Block.terminateDrag_`?

joshlory: I definitely can if you prefer.  Is it better to call `Block.terminateDrag_` directly?  It looks like `BlockSpaceEditor.terminateDrag_` also calls `Flyout.terminateDrag_`:

```js
Blockly.BlockSpaceEditor.terminateDrag_ = function() {
  Blockly.Block.terminateDrag_();
  Blockly.Flyout.terminateDrag_();
};
```

Hamms: I'm so upset that we're calling a private function on a separate class which itself calls a private function on two other separate classes. :upside_down_face: 

It's probably better in that case to call the BlockSpaceEditor version.

PR #42, opened 2017-04-11
New speculative fix for `this.sourceBlock_ is null` on all field types

joshlory:
PR https://github.com/code-dot-org/blockly/pull/38 indeed got rid of the `Blockly.FieldDropdown` error, however we started seeing new errors for `Blockly.FieldTextInput`.  Rather than add the null check for each field check, instead add it to the scroll listener before `this.positionWidgetDiv()` is called.

### DIFF
--- a/core/ui/fields/field.js
+++ b/core/ui/fields/field.js
@@ -379,6 +379,12 @@ Blockly.Field.prototype.positionWidgetDiv = function() {
   // intentional noop; overridden by child classes
 };
 
+Blockly.Field.prototype.handleBlockSpaceScrolled = function () {
+  if (this.sourceBlock_) {
+    this.positionWidgetDiv();
+  }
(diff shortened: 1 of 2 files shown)


core/ui/fields/field.js
```diff
@@ -379,6 +379,10 @@ Blockly.Field.prototype.positionWidgetDiv = function() {
   // intentional noop; overridden by child classes
 };
 
+Blockly.Field.prototype.handleBlockSpaceScrolled = function () {
+  this.positionWidgetDiv();
```
Hamms: is this supposed to have the `if (!this.sourceBlock_) {` check?

joshlory: 😳  yes.

joshlory:
PTAL.

PR #45, opened 2017-04-21
Remove an un-needed assert in block limits

caleybrock:
Causing an assertion error in New Relic when levelbuilders make the limit smaller that the start block count. 
`AssertionError: Failure: this toolbox block cannot create more than 0 workspace blocks`

Elijah suggested just removing because the alert doesn't do anything useful, since this is a valid use case.

cc @joshlory


joshlory:
This looks good, but I think @ryansloan and I were leaning towards option (B) here: https://codedotorg.slack.com/archives/C0T0GC9GE/p1491520376103128

> Change when we check the assert so it only happens when dragging blocks from the toolbox.

I think the assert is still valuable to identify when levels are misconfigured.  The level that's throwing the `AssertionError` is a special case because we want the student to delete the extra block to fix the puzzle, without making it available in the toolbox.

Hamms:
note that we'd also have to have it happen when copy/pasting blocks; really, anytime a block is being _added_, just not when one is being _removed_.

Are we certain we'd actually use this to identify misconfigured levels? I feel like it would end up getting ignored.

caleybrock:
I'm on the same page as @Hamms but understand if you're looking for something else. If there's more to be done, I'll probably have to bump this item until at least next week.

joshlory:
Should we go ahead and remove the alert until @ryansloan specs out how limited toolbox blocks should work?  There's also the question of whether block counts should be allowed to go negative (to let students drag out repeated blocks before collapsing them into a `repeat` loop).

caleybrock:
Seems reasonable to me.

caleybrock:
Sorry this took so long! @Hamms let me know if you think it's worth publishing for just this or just waiting for the next change.

Hamms:
I'm actually just about to do an i18n sync which usually requires an blockly update, so I'll bundle this in with that!

PR #53, opened 2017-06-21
allowing drop down menus to reposition near the edge of workspace

WilliamHarmonMS:


joshlory:
@Hamms do you mind taking a look at this PR?  It's for large dropdown menus in Minecraft Codebuilder.

William, can you add screenshots to show how this feature works?  Also worth testing in a RTL language.

Thanks!

WilliamHarmonMS:
Sure can! Ended up being a pretty small adjustment with a bit of math taking into account workspace dimensions. Attachments to follow. In order:
-Normal Behavior
-Problem case
-Case0 fix
-Case1 fix
-Case0 + Case1 fix
normal behavior
problem behavior
right edge fix
bottom edge fix
bottom-right corner fix

core/ui/fields/field_rectangular_dropdown.js
```diff
@@ -266,12 +266,22 @@ Blockly.FieldRectangularDropdown.prototype.addPositionAndShowMenu = function (me
  * @override
  */
 Blockly.FieldRectangularDropdown.prototype.positionWidgetDiv = function () {
+  var width = this.menu_.element_.clientWidth;
+  var height = this.menu_.element_.clientHeight;
```
Hamms: nit: this should be using `this.menu_.getElement()`, and ideally also `goog.style.getSize`

core/ui/fields/field_rectangular_dropdown.js
```diff
@@ -266,12 +266,22 @@ Blockly.FieldRectangularDropdown.prototype.addPositionAndShowMenu = function (me
...
 
   var menuPosition = this.calculateMenuPosition_(this.previewElement_, positionBelow);
 
   var windowSize = goog.dom.getViewportSize();
+  
+  if(menuPosition.x + width > windowSize.width) {
+  	menuPosition.x -= width - (width / numberOfColumns);
+  }
+  if(menuPosition.y + height > windowSize.height) {
+  	menuPosition.y -= height + (height / Math.ceil(this.menu_.getChildCount() / numberOfColumns));
```
Hamms: it's not clear to me why we're considering number of columns in our height calculation

core/ui/fields/field_rectangular_dropdown.js
```diff
@@ -266,12 +266,22 @@ Blockly.FieldRectangularDropdown.prototype.addPositionAndShowMenu = function (me
...
+  var width = this.menu_.element_.clientWidth;
+  var height = this.menu_.element_.clientHeight;
   var numberOfColumns = chooseNumberOfColumns(this.menu_.getChildCount());
   var positionBelow = numberOfColumns > 1;
 
   var menuPosition = this.calculateMenuPosition_(this.previewElement_, positionBelow);
 
   var windowSize = goog.dom.getViewportSize();
+  
+  if(menuPosition.x + width > windowSize.width) {
```
Hamms: I believe both this and the height check should take `scrollOffset` into account

core/ui/fields/field_rectangular_dropdown.js
WilliamHarmonMS: It's to correctly space the drop down menu. For instance with the drop down menu I'm working with, 116 items -> 10.77 columns truncated to 10. What Math.ceil(this.menu_.getChildCount() / numberOfColumns) gets us is 116/10 -> 11.6 -> ceilinged to 12, which is giving up the row count (because at 10 items a row, 10 columns, that leaves the last row to cover items 111 through 120, which is half full at 116)

In essence I bumped the drop down 1 extra row up, position wise. If you look at the top image, top left corner of the drop down aligned with bottom left of the "button". If you bumped it up merely by the height, bottom left matches with bottom left and covers the button. The extra bump moves bottom left corner of dropdown to align with top left corner of button as seen in image 4.

Is this ok?

core/ui/fields/field_rectangular_dropdown.js
WilliamHarmonMS: Roger that, will fix now.

core/ui/fields/field_rectangular_dropdown.js
WilliamHarmonMS: I can see this for the height check. However given that the workspace doesn't scroll horizontally, does that apply to the width check as well?

core/ui/fields/field_rectangular_dropdown.js
WilliamHarmonMS: quick question about this. In my case specifically:
this.menu_.getElement().clientWidth = 364
goog.style.getSize(this.menu_.getElement()) returns a goog.math.Size that has a width of 366.

Why the discrepancy, and which should I use? It likely won't make a huge difference for -my- case, but I worry that in other cases this discrepancy might be larger?

core/ui/fields/field_rectangular_dropdown.js
WilliamHarmonMS: Also: In testing this out, my scrollOffset was (0,0) in every case, even when extending the workspace. In what case is the scrollOffset non-0? I'm trepid about adding things to the code that I can't test and verify.

core/ui/fields/field_rectangular_dropdown.js
Hamms: `getSize` returns the height and width of the border box, regardless of which box model is in effect, while `clientWidth` will return either the content or border box. A diff of 2px is 100% expected!

WilliamHarmonMS: Good! Since this is my first foray into this specific code base, I wanted to make sure I wasn't missing something nuanced. Already pushed the change to use goog.style.getSize(this.menu_.getElement())

core/ui/fields/field_rectangular_dropdown.js
Hamms: oh, I guess I misunderstood `scrollOffset`; it's actually _page_ scroll, not _workspace_ scroll, which for our purposes is irrelevant.

So, nevermind!

WilliamHarmonMS: Was about to edit that comment that it seemed like it was a page scroll and impossible via code.org in its current state. I don't know if these changes propagate to using these tools in some other environment though, so I definitely wanted to double check. Would just be hard for me to test. Glad to hear this is alright though.

core/ui/fields/field_rectangular_dropdown.js
Hamms: Oh, sorry, I should have been more clear. I mean why are we using the content of the dropdown to determine the offset rather than the height of the block.

That being said, this seems to work fine, so I don't have a particularly strong preference.

WilliamHarmonMS: I was moving it up the width of a single row, so that it appeared just above the button. 2 rows means 1/2 the window height, 10 rows means 1/10th the window height, so on and so forth. I mean an easier way would probably be "the height of 1 element" if that data lives somewhere (though I don't know how that might be stored differently if the elements are text vs images).

End of the day I resort to math when I can't find things in code bases I'm unfamiliar with. Haha. Though if it works to your liking, I'm happy with it.

joshlory:
@Hamms merge this?

WilliamHarmonMS:
::whispers in the distance:: ~merrrrrrrrrge it~

Hamms:
Oh, whoops. My bad for completely forgetting about this!

PR #55, opened 2017-07-07
Add disable autopopulate option

Hamms:
As it currently stands, Blockly works some magic with Function categories. Specifically, when it detects that a category is supposed to be a Function category, it ignores any levelbuilder-specified XML for that category and instead populates it with three different function definition blocks; `noreturn`, `return`, and `ifreturn`. It also populates it with call blocks for any functions that have already been defined.

Unfortunately, there are situations in which we want to present students with a subset of those definition blocks (mostly situations in which we have not yet introduced the concept of returns and would like to only expose `noreturn`). We therefore would like a new Blockly option for disabling the autopopulation of function definitions and instead actually relying on the levelbuilder-specified XML.

### DIFF
--- a/core/initialization/inject.js
+++ b/core/initialization/inject.js
@@ -159,6 +159,7 @@ Blockly.parseOptions_ = function(options) {
     disableIfElseEditing: options['disableIfElseEditing'] || false,
     disableParamEditing: options['disableParamEditing'] || false,
     disableVariableEditing: options['disableVariableEditing'] || false,
+    disableProcedureAutopopulate: options['disableProcedureAutopopulate'] || false,
     useModalFunctionEditor: options['useModalFunctionEditor'] || false,
     useContractEditor: options['useContractEditor'] || false,
     disableExamples: options['disableExamples'] || false,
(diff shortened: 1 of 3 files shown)


Hamms:
FYI @kiki-lee

joshlory:
Should this be the default behavior?  (I.e. not make it a configurable option, but implement this for all Function categories that use the legacy function definition blocks.)

kiki-lee:
Default, yes, but still maybe overridable?

On Jul 7, 2017 4:22 PM, "Josh Lory" <notifications@github.com> wrote:

Should this be the default behavior? (I.e. not make it a configurable
option, but implement this for all Function categories that use the legacy
function definition blocks.)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/code-dot-org/blockly/pull/55#issuecomment-313814943>,
or mute the thread

Hamms:
We'd want to do a pass on literally every level that's using categories to make sure that they have the desired manual XML before switching the defaults, otherwise we could end up with a bunch of levels that have no way to define functions.

ryansloan:
A couple of thoughts:
1) We'll continue to use the non-modal function editor in our curriculum forever. There are points when we are introducing these things to students in a lightweight way where they aren't ready for the out-of-sight level of abstraction the modal provides.
2) To my knowledge, we _never_ use the return-related blocks, right @kiki-lee ? I don't know of any instance in our K-5 curriculum or HOC where this is used.

PR #61, opened 2017-08-24
Add angle helper to value inputs

Hamms:
Teach the existing FieldTextInput ow to leverage the new Field Helpers to display an Angle Helper when enabled.

This mostly involves moving a bunch of functionality from the child class FieldAngleTextInput into its parent; the child is now simply a version of FieldTextInput for which the angle helper is always enabled, rather than conditionally.

### DIFF
--- a/core/ui/fields/field_angle_textinput.js
+++ b/core/ui/fields/field_angle_textinput.js
@@ -1,70 +1,35 @@
 /**
- * @fileoverview Text input field.
- * @author fraser@google.com (Neil Fraser)
+ * @fileoverview special Text Input field that always shows angle helper
+ * @author elijah@code.org (Elijah Hamovitz)
  */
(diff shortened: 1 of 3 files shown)


core/ui/fields/field_angle_textinput.js
```diff
@@ -1,70 +1,35 @@
...
 goog.provide('Blockly.FieldAngleTextInput');
-goog.require('Blockly.FieldTextInput');
+
 goog.require('Blockly.AngleHelper');
+goog.require('Blockly.BlockFieldHelper');
+goog.require('Blockly.FieldTextInput');
 
 /**
- * Class for an editable text field.
+ * Class for an editable text field which will always show the Angle HElper
```
balderdash: nit: s/HElper/Helper/

PR #68, opened 2017-11-03
Autopopulate function call blocks in uncategorized toolbox

balderdash:
screenshot from 2017-11-03 11-33-53

This PR adds a `topLevelProcedureAutopopulate` blockly config option, which when enabled adds function call blocks to the categoryless toolbox for every function defined in the workspace. If the toolbox has categories, this option has no effect.


core/ui/block_space/flyout.js
```diff
@@ -455,10 +456,16 @@ Blockly.Flyout.prototype.show = function(xmlList) {
...
 
-    Blockly.Procedures.flyoutCategory(blocks, gaps, margin,
-      this.blockSpace_,
-      function(procedureInfo) { return !procedureInfo.isFunctionalVariable; }
-    );
+    if (Blockly.topLevelProcedureAutopopulate) {
+      this.layoutXmlToBlocks_(xmlList, blocks, gaps, margin);
+    }
+
+    if (Blockly.mainBlockSpace) {
```
Hamms: why is this dependent on `mainBlockSpace`?

core/ui/block_space/flyout.js
```diff
@@ -744,6 +751,16 @@ Blockly.Flyout.prototype.createBlockFunc_ = function(originBlock) {
...
+    if (Blockly.topLevelProcedureAutopopulate && block.isFunctionDefinition()) {
+      block.blockEvents.listenOnce(
+        Blockly.Block.EVENTS.AFTER_DROPPED,
+        function() {
+          window.setTimeout(function() {
+            targetBlockSpace.blockSpaceEditor.updateFlyout();
+          }, 0);
+        }
+      );
+    }
```
Hamms: why is there not a comparable method for updating the flyout when in categories mode?

core/ui/block_space/block_space_editor.js
```diff
@@ -519,6 +519,20 @@ Blockly.BlockSpaceEditor.prototype.init_ = function() {
   }
 };
 
+Blockly.BlockSpaceEditor.prototype.updateFlyout = function() {
+  if (this.flyout_ && Blockly.languageTree) {
+    this.flyout_.show(Blockly.languageTree.childNodes);
+    this.svgResize();
```
Hamms: :+1:

core/ui/block_space/flyout.js
Hamms: This also appears to be duplicated in [block.onMouseUp](https://github.com/code-dot-org/blockly/pull/68/files#diff-5a399674d5ba4dd56af39c1c8df3fe22R801). Is that intentional/necessary?

core/ui/block_space/flyout.js
balderdash: `Blockly.Procedures.flyoutCategory` finds all the procedure definitions in `mainBlockSpace`, and blows up if it's not defined.

core/ui/block_space/flyout.js
balderdash: When you're in categories mode, the flyout closes as soon as you start adding/removing/renaming blocks in the main blockspace. It then gets recreated from scratch every time you open the function category, so there's no need to update it immediately after an edit.

Yeah, I need both. The `onMouseUp` handler is only for when you're dropping a block in the trashcan, and this one is only for when you first drag a block out of the toolbox.

PR #70, opened 2017-11-11
Update blockly_compressed.js

balderdash:
Forgot to rebuild in non-debug mode before merging https://github.com/code-dot-org/blockly/pull/68


Hamms:
That's actually not necessary; both builds will get made and added automatically when creating a new release version

Hamms:
See https://github.com/code-dot-org/blockly/blob/staging/package.json#L21-L23

PR #71, opened 2017-11-15
Don't bind dom events on temporary blockspaces

balderdash:
This fixes the workspace size issue that was still present on mobile safari on minecraft levels.

Handlers for DOM events are only set up for the first blockspace that gets created. Usually, the main workspace is that blockspace and everything is fine, but that changes when initialization blocks are present. The `xmlToBlocks` method creates a temporary blockspace, which steals the event handlers, and breaks resizing.

There's actually nothing specific to mobile safari about this, and it's almost completely unrelated to the very similar-looking bug I fixed last week in https://github.com/code-dot-org/code-dot-org/pull/18937. This bug is present for all browsers, for any level with initialization blocks, it's just way more noticeable on mobile safari because it automatically resizes right after loading the page.

My fix is to skip setting up the DOM event handlers for the temporary blockspace used in `xmlToBlocks`, allowing the actual workspace to get them instead.

### DIFF
--- a/core/ui/block_space/block_space_editor.js
+++ b/core/ui/block_space/block_space_editor.js
@@ -70,6 +70,9 @@ Blockly.BlockSpaceEditor = function(container, opt_options) {
   if (opt_options.disableTooltip) {
     this.disableTooltip = opt_options.disableTooltip;
   }
+  if (opt_options.disableEventBindings) {
+    this.disableEventBindings = opt_options.disableEventBindings;
+  }
 
   this.readOnly_ = !!opt_options.readOnly;
   this.noScrolling_ = !!opt_options.noScrolling;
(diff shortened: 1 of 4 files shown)


core/ui/block_space/block_space_editor.js
```diff
@@ -89,7 +91,7 @@ Blockly.BlockSpaceEditor = function(container, opt_options) {
   this.blockLimits = new Blockly.BlockLimits();
 
   this.createDom_(container);
-  this.init_();
+  this.init_(!opt_options.disableEventBindings);
```
Hamms: any reason why this option isn't just getting saved as an instance attribute like the others?

balderdash: Oh, that does make more sense. Done.

PR #75, opened 2017-12-06
Allow limited blocks to go below zero, add warning icon

joshlory:
Request from @ryansloan.

Spec: https://docs.google.com/document/d/1HAiICw2ae3644U6PrU3oX5zaibNvNqkbyPOBXuf-F8M/edit#heading=h.758y0aszrrbe

Before:

limited-before

After:

limited-after


joshlory:
Yep!  The warning will look like this, but will also show the block that there is too many of (along with its limit).

screen shot 2017-12-13 at 5 29 59 pm

joshlory:
@Hamms here is the WIP diff: https://github.com/code-dot-org/code-dot-org/compare/staging...limited-block-improvements

PR #83, opened 2018-03-14
Update blockly to use Circle 2.0

balderdash:
Mostly just used the auto-generated config and removed the parts we didn't need.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,12 +9,10 @@ jobs:
     - checkout
     - restore_cache:
         keys:
-        - v2-dep-{{ .Branch }}-
-        - v2-dep-staging-
-        - v2-dep-
+        - v3-dep-{{ checksum "package.json" }}
     - run: npm install
     - save_cache:
-        key: v1-dep-{{ .Branch }}-{{ epoch }}
+        key: v3-dep-{{ checksum "package.json" }}
         paths:
         - ./node_modules
     - run: npm test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,20 @@
+version: 2
+jobs:
+  build:
+    working_directory: ~/code-dot-org/blockly
+    parallelism: 1
+    docker:
+    - image: circleci/node:6.13
```
balderdash: Circle offers a set of docker images, but they didn't have one at node 6.9, only 6.13 (see: https://hub.docker.com/r/circleci/node/tags/).

We could create our own image like Will did for the main repo to get the right version, but I'm not sure that's worth the effort.

Hamms: Agreed; 6.13 should be fine

.circleci/config.yml
```diff
@@ -0,0 +1,20 @@
+version: 2
+jobs:
+  build:
+    working_directory: ~/code-dot-org/blockly
```
Hamms: why not just `blockly`?

.circleci/config.yml
```diff
@@ -0,0 +1,20 @@
...
+    parallelism: 1
+    docker:
+    - image: circleci/node:6.13
+    steps:
+    - checkout
+    - restore_cache:
+        keys:
+        - v2-dep-{{ .Branch }}-
```
Hamms: should this also have `{{ epoch }}` like `save_cache`?

Hamms: and actually, shouldn't both be using something like `checksum "package.json"`?

balderdash:
Yup! The stripped-down container loads quicker than the old circle 1.0 environment.

PR #85, opened 2018-03-16
Use a curve that's actually 50px tall to check for bbox bug

balderdash:
Followup to https://github.com/code-dot-org/blockly/pull/7

Chrome has since fixed https://bugs.chromium.org/p/chromium/issues/detail?id=230599, but our bug-detector is giving us false positives. The current test path actually does extend into negative-y terroritory, so chrome correctly reports it as having a height greater than 50: 
image
The triangle has a height and width of 50, the little notch at the top left is the bezier curve you get by going from (0,0) to (0,0) with both control points at (0, -5)

I'm replacing it with a curve that extends from (0, 50) to (125, 0), with the first control point at (75, -25). This catches older versions of chrome that include the control point, and thus report the height as 75px, but not newer ones that correctly determine that the curve stays on the sunny side of the x-axis:
image

### DIFF
--- a/core/ui/block_space/block_space_editor.js
+++ b/core/ui/block_space/block_space_editor.js
@@ -558,7 +558,7 @@ Blockly.BlockSpaceEditor.prototype.detectBrokenControlPoints = function() {
      */
     var container = Blockly.createSvgElement('g', {}, this.svg_);
     Blockly.createSvgElement('path',
-      {'d': 'm 0,0 c 0,-5 0,-5 0,0 H 50 V 50 z'}, container);
+      {'d': 'M 0,50 C 75,-25 75,50 125,0 Z'}, container);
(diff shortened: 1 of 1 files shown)


balderdash:
This also fixes a bug in our block highlighting that's been bugging me for quite some time. Note the notches in the corners.
Before:
image

After:
image

balderdash:
According to the chromium bug page, Chrome 62 and older. I spot-checked chrome 60 and saw that it was still present.

Hamms:
Might be nice to add a note in there saying something to the effect of "if at any point we remove support for Chrome version 62 and older, this bug detection can be removed"

balderdash:
Turns out this also fixed another bug that has been annoying me:
Before:
image

After:
image

PR #86, opened 2018-03-19
 Add an option to strictly enforce connection types

balderdash:
Input connections of type "None" will accept any type of output connection, and output connections of type "None" can connect to any input. At the moment, blockly only stops you from making a connection if both sides specify a type and the types don't match.

This adds a `strictType_` property to connections in case you ever need to reject connections to an untyped connection (say, if you want to make a sprite block with a connector that has a different shape). 

This gif shows the new green sprite variable blocks and how they reject connections to normal variable blocks because the sprite versions have strict type checking:

connections2


balderdash:
Yeah that makes sense.

I defined them in blockly to make testing easier, but I can move them to `playground_requires.js` for that.

PR #89, opened 2018-03-21
Triangular block tabs for sprite-typed connections

balderdash:
angle tabs

### DIFF
--- a/core/ui/block_svg/block_svg.js
+++ b/core/ui/block_svg/block_svg.js
@@ -239,7 +239,7 @@ BS.TAB_PATH_DOWN_HIGHLIGHT_RTL = 'v 6.5 m -' +
     (BS.TAB_WIDTH * 0.98) + ',2.5 q -' +
     (BS.TAB_WIDTH * .05) + ',10 ' +
     (BS.TAB_WIDTH * .27) + ',10 m ' +
-    (BS.TAB_WIDTH * .71) + ',-2.5 v 1.5';
+    (BS.TAB_WIDTH * .71) + ',-2.5 v 3.5';
 
 /**
  * SVG paths for drawing a horizontal puzzle tab from bottom to top.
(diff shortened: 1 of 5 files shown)


ryansloan:
Sweet! I think this is great for our classroom tests, but if we find that the tab alone is hard to distinguish from the standard rounded tab, would it be straightforward for us to make the right side triangular as well?

balderdash:
It's not exactly straightforward to change the shape of the right side. What it would look like if you had a block that took a non-inlined input but output a sprite? Like this, but where `not` returns a sprite:
image

We could try to ensure that blocks that can output a sprite only have inline inputs, but that feels very limiting.

joshlory:
Ah, good point.  Are there cases where we'd want to operate on sprites, and return a sprite?

The main case I can think of is groups.  My preference would be to treat sprites and groups the same, so a single sprite is actually a sprite group with only one entry.  Thoughts?

balderdash:
Yeah making sprites and groups work wherever the other can be used was what I had in mind, so a block like `get sprite N from group GROUP` would take a "sprite" as input and also output a sprite.

We could maybe do some wizardry to add a triangle tab to the right side of an input if it gets plugged into a block with non-inline input and a sprite output, but that breaks down (or at least gets _incredibly_ hairy) if you have multiple rows of inputs.

core/ui/block_svg/block_svg.js
```diff
@@ -233,7 +241,92 @@ BS.TAB_PATH_DOWN_HIGHLIGHT_RTL = 'v 6.5 m -' +
...
+ */
+BS.ANGLE_TAB_PATH_DOWN =
+  'v ' + (BS.TAB_HEIGHT / 3) + ' ' +
+  'l ' + (BS.TAB_WIDTH * -1) + ' ' + (BS.TAB_HEIGHT / 3) + ' ' +
+  'l ' + (BS.TAB_WIDTH) + ' ' + (BS.TAB_HEIGHT / 3);
+
+BS.ANGLE_TAB_PATH_DOWN_HIGHLIGHT_RTL =
+  'v 5.5 ' +
+  'm -' + (BS.TAB_WIDTH - 1) + ',' + (BS.TAB_HEIGHT / 3 + 1) + ' ' +
+  'l ' + (BS.TAB_WIDTH - 1) + ',' + (BS.TAB_HEIGHT / 3);
```
joshlory: Worth following the same pattern as PR https://github.com/code-dot-org/blockly/pull/84 for RTL mode?

balderdash: Highlights aren't mirrored when you switch to RTL. The light source is always meant to come from the upper left, so different parts of the tab get a highlight.

image
image

balderdash: Er, there's two uses of "highlight". This highlight is the fake reflection from the light source. The highlight in https://github.com/code-dot-org/blockly/pull/84 is the bright yellow highlight you see when you're about to connect two blocks.

PR #91, opened 2018-03-22
Only count used blocks towards the toolbox block limits

joshlory:
Fixes https://codeorg.zendesk.com/agent/tickets/137867.

Unfortunately we don't have any existing test coverage for `flyout.js`.  I'll get this tracked on the tech debt backlog.


Hamms:
Will this mean that the block limit indicator won't update until the block gets attached, rather than updating when it gets "pulled" from the toolbox?

PR #92, opened 2018-03-22
Cleanup: we no longer disable blocks that are over their toolbox limits

joshlory:
I should've removed this in PR https://github.com/code-dot-org/blockly/pull/75.


joshlory:
I thought this was for toolbox block limits, but it's actually for the global `maxBlocks` setting. 
 Abandoning.

PR #93, opened 2018-04-02
Bug fix: only count user-visible blocks

joshlory:
Fix for PR https://github.com/code-dot-org/blockly/pull/91.


joshlory:
Good call, updated.

PR #100, opened 2018-04-21
Type hints

joshlory:
Try out adding small visual hints for what type of block a connection accepts.  The type data is already there for a lot of blocks, we just don't expose that information visually.  This option is off by default.

image


core/connections/connection.js
```diff
@@ -420,15 +420,41 @@ Blockly.Connection.prototype.moveBy = function(dx, dy) {
...
+    case 'String':
+      return '#5ba68d';
+    case 'Number':
+      return '#77679e';
+    case 'Colour':
+      return '#0094ca';
+    case 'Sprite':
+      return '#399e4b';
+    case 'Behavior':
+      return '#d1c404';
```
Hamms: are these type-color pairings also defined anywhere else?

core/ui/block_svg/block_svg.js
```diff
@@ -69,6 +69,16 @@ Blockly.BlockSvg.prototype.initChildren = function () {
     'transform': 'translate(1, 1)',
     'fill-rule': 'evenodd'
   }, this.svgGroup_);
+  if (Blockly.typeHints) {
+    this.svgTypeHints_ = Blockly.createSvgElement('g', {
+      'class': 'blocklyTypeHint'
+    }, this.svgGroup_);
+    for (var i = 0; i < this.block_.inputList.length; i++) {
+      Blockly.createSvgElement('path', {
```
Hamms: any specific reason why these paths are created here and modified in `renderDraw` rather than just doing everything in `renderDraw`?

core/connections/connection.js
```diff
@@ -420,15 +420,41 @@ Blockly.Connection.prototype.moveBy = function(dx, dy) {
...
- * Add highlighting around this connection.
+ * Get the hint type color for a given set of strict checks.
  */
-Blockly.Connection.prototype.highlight = function() {
+Blockly.Connection.prototype.colorForType = function(checks) {
+  if (!checks) {
+    return;
+  }
+  if (checks.length > 1) {
+    //throw new Error('Cannot show type hints for multiple strict types.');
```
Hamms: should this be uncommented?

core/ui/block_space/block_space_editor.js
```diff
@@ -231,6 +231,18 @@ Blockly.BlockSpaceEditor.prototype.populateSVGEffects_ = function(container) {
...
+   <filter id="blocklyTypeHintFilter">
+     <feConvolveMatrix order="5,5" kernelMatrix="0 0 ... 0.5 1">
+   </filter>
+   */
+  filter = Blockly.createSvgElement('filter', {
+    id: 'blocklyTypeHintFilter'
+  }, defs);
+  Blockly.createSvgElement('feConvolveMatrix', {
+    order: '5,5',
+    kernelMatrix: '0 0 0 0 0 0 0 0 0.5 1 0.5 1 1 1 1 0 0 0 0 0 0 0 0 0.5 1'
```
Hamms: can you confirm that this works in safari? [MDN is uncertain](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/feConvolveMatrix#Browser_compatibility)

core/ui/block_svg/block_svg.js
joshlory: I think the idea is to create the DOM elements once, then only modify them in `renderDraw`.  Not 100%, I was following the existing pattern here.

core/connections/connection.js
joshlory: Not yet 🙁.  I _think_ this should be ok for the experiment.  If we decide to make this feature more available we may want to look at automatic block colors (overridable in `init`) that match the return type.

core/ui/block_space/block_space_editor.js
joshlory: Confirmed in Safari.  It looks a little bit different:

image

PR #101, opened 2018-04-21
Don't throw for unknown language blocks

joshlory:
Given the following snippet, containing unknown language blocks `when_run` and `draw_move_by_constant_dropdown`:

```js
startBlocks = `<xml>
  <block type="when_run" deletable="false" movable="false">
    <next>
      <block type="draw_move_by_constant_dropdown">
        <title name="DIR">moveForward</title>
        <title name="VALUE" config="50,100,150,200,300">100</title>
      </block>
    </next>
  </block>
</xml>`;
Blockly.mainBlockSpace.clear();
var dom = Blockly.Xml.textToDom(startBlocks);
Blockly.Xml.domToBlockSpace(Blockly.mainBlockSpace, dom);
```

### Before:

_Empty blockspace._

### After:

<image width="336" src="https://user-images.githubusercontent.com/413693/39089722-881480fe-4582-11e8-8bc7-947970cd9c0c.png">


core/ui/block.js
```diff
@@ -124,7 +124,9 @@ Blockly.Block = function(blockSpace, prototypeName, htmlId) {
     this.type = prototypeName;
     var prototype = Blockly.Blocks[prototypeName];
     if (!prototype) {
-      throw 'Error: "' + prototypeName + '" is an unknown language block.';
+      prototype = Blockly.Blocks.unknown;
+      this.appendDummyInput().appendTitle('unknown: ' + prototypeName);
+      console.warn('Warning: "' + prototypeName + '" is an unknown language block.');
```
balderdash: Console warnings don't get reported to newrelic, right? Do we want to instrument these somehow?

joshlory: Maybe at some point?  Another way to get the information about (unknown) block usage would be to run Redshift queries on the level sources table.

I could get rid of the warning if you don't think it's useful.

balderdash: The warning is still useful at least during development.

I'm just worried that we'll break a block and not know about it until we get zendesk reports, whereas today the resulting errors would show up immediately in newrelic.

joshlory: Ah that's a good point.  How should I connect up those errors?  I can't log directly to New Relic or Firehose from inside Blockly, but I could expose an event that the consumer could add a listener for?

```js
Blockly.fireUiEvent(window, 'unknownBlock', {name: prototypeName});
```

joshlory:
[Don't throw for unknown language blocks](https://trello.com/c/pIakwqvd/112-dont-throw-for-unknown-language-blocks)

core/ui/block.js
balderdash: Exposing an event SGTM.

joshlory:
screen shot 2018-04-27 at 11 29 43 am

screen shot 2018-04-27 at 11 29 24 am

joshlory:
@balderdash PTAL.

PR #103, opened 2018-04-24
Delete bld_flags.txt

joshlory:
Remove unused build flags file.


Hamms:
I believe this file is used by the build script: https://github.com/code-dot-org/blockly/blob/staging/deploy.sh#L24

joshlory:
Whoop — closing.

PR #104, opened 2018-05-07
Add variable categories

balderdash:
In the process of creating the behavior editor, I've come across the need to create a toolbox category that autopopulates with a separate category of variable getter blocks (in particular for behaviors. but this is also useful for sprite and location variables). This PR enables that.

Basically, this PR modifies every blocks' `getVars` method to return a map of variables by category (usually there's just one category per block), so that the flyout creation code can find all the variables of a given category. All existing variables are now in the "Default" category, and https://github.com/code-dot-org/code-dot-org/pull/22233 puts the sprite variable getter/setter blocks into the "Sprite" category.

### DIFF
--- a/blocks/functional_procedures.js
+++ b/blocks/functional_procedures.js
@@ -211,11 +211,10 @@ Blockly.Blocks.functional_definition = {
       isFunctionalVariable: this.isFunctionalVariable_
     }
   },
-  getVars: function(category) {
-    if (category && category !== Blockly.Variables.DEFAULT_CATEGORY) {
-      return [];
-    }
-    return this.parameterNames_;
+  getVars: function() {
+    return {
+      Default: this.parameterNames_,
+    };
   },
   renameVar: function(oldName, newName) {
     var change = false;
--- a/blocks/functional_procedures.js
+++ b/blocks/functional_procedures.js
@@ -212,7 +212,7 @@ Blockly.Blocks.functional_definition = {
     }
   },
   getVars: function(category) {
-    if (category) {
+    if (category && category !== Blockly.Variables.DEFAULT_CATEGORY) {
       return [];
     }
     return this.parameterNames_;
--- a/blocks/functional_procedures.js
+++ b/blocks/functional_procedures.js
@@ -211,7 +211,10 @@ Blockly.Blocks.functional_definition = {
       isFunctionalVariable: this.isFunctionalVariable_
     }
   },
-  getVars: function() {
+  getVars: function(category) {
+    if (category) {
+      return [];
+    }
     return this.parameterNames_;
   },
   renameVar: function(oldName, newName) {
--- a/blocks/loops.js
+++ b/blocks/loops.js
@@ -117,7 +117,7 @@ Blockly.Blocks.controls_for = {
     });
   },
   getVars: function(category) {
-    if (category) {
+    if (category && category !== Blockly.Variables.DEFAULT_CATEGORY) {
       return [];
     }
     return [this.getTitleValue('VAR')];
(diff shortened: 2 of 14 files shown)


blocks/functional_procedures.js
```diff
@@ -211,7 +211,10 @@ Blockly.Blocks.functional_definition = {
       isFunctionalVariable: this.isFunctionalVariable_
     }
   },
-  getVars: function() {
+  getVars: function(category) {
+    if (category && category !== Blockly.Variables.DEFAULT_CATEGORY) {
+      return [];
+    }
```
joshlory: Can we share this code?  (repeated 7x below)

blocks/loops.js
```diff
@@ -116,7 +116,10 @@ Blockly.Blocks.controls_for = {
           thisBlock.getTitleValue('VAR'));
     });
   },
-  getVars: function() {
+  getVars: function(category) {
+    if (category && category !== Blockly.Variables.DEFAULT_CATEGORY) {
+      return [];
+    }
     return [this.getTitleValue('VAR')];
```
joshlory: Can we share this code?  (repeated 7x)

blocks/functional_procedures.js
```diff
@@ -212,7 +212,9 @@ Blockly.Blocks.functional_definition = {
     }
   },
   getVars: function() {
-    return this.parameterNames_;
+    return {
+      Default: this.parameterNames_,
```
Hamms: could this be `[Blockly.Variables.DEFAULT_CATEGORY]: this.parameterNames_` or does blockly not support that?

core/utils/variables.js
```diff
@@ -39,13 +39,22 @@ goog.require('Blockly.BlockSpace');
...
  * @param {Array.<Blockly.Block>=} opt_blocks Optional root blocks.
+ * @param {string=} opt_category only return variables in this category, or all
+ *   variables if not specified
  * @return {!Array.<string>} Array of variable names.
  */
-Blockly.Variables.allVariables = function(opt_blocks) {
+Blockly.Variables.allVariables = function(opt_blocks, opt_category) {
+  if (opt_category && opt_category !== Blockly.Variables.DEFAULT_CATEGORY &&
+      (Blockly.valueTypeTabShapeMap &&
+      Object.keys(Blockly.valueTypeTabShapeMap).indexOf(opt_category) === -1)) {
```
Hamms: why not `Blockly.valueTypeTabShapeMap[opt_category] !== undefined`?

core/utils/variables.js
```diff
@@ -81,6 +94,32 @@ Blockly.Variables.allVariables = function(opt_blocks) {
...
+ * @returns {string[]} Array of all the variables used.
+ */
+Blockly.Variables.allVariablesFromBlock = function(block) {
+  if (!block.getVars) {
+    return [];
+  }
+  var varsByCategory = block.getVars();
+  return Object.keys(varsByCategory).reduce(function (vars, category) {
+    return vars.concat(varsByCategory[category]);
+  }, []);
```
Hamms: :+1:

core/utils/variables.js
```diff
@@ -170,6 +218,30 @@ Blockly.Variables.flyoutCategory = function(blocks, gaps, margin, blockSpace) {
...
+  'Default': 'variables_set',
+};
+Blockly.Variables.getSetter = function(blockSpace, category) {
+  var setterName = Blockly.Variables.setters[category];
+  return (setterName && Blockly.Blocks[setterName]) ?
+      new Blockly.Block(blockSpace, setterName) : null;
+};
+Blockly.Variables.registerSetter = function(category, blockName) {
+  Blockly.Variables.setters[category] = blockName;
+};
```
Hamms: this is delightful

blocks/functional_procedures.js
balderdash: PhantomJS mysteriously times out running blockly tests when I try to use that syntax. I took that to mean it doesn't support it.

core/utils/variables.js
balderdash: Because I did not think of that.

PR #105, opened 2018-05-10
Make the function editor more flexible so that it can become a behavior editor

balderdash:
Adds a few options to the function editor so that, with the right constructor arguments, it can be a behavior editor instead.
* `msgOverrides` a map a of strings to replace various UI strings, e.g. replacing "Function" with "Behavior"
* `definitionBlockType` which block type to use for function definitions. Defaults to `procedures_defnoreturn`
* `opt_parameterBlockTypes` a map of block value types to the corresponding parameter blocks
* `opt_disableParamEditing` disables param editing for just this function editor, rather than globally.

See https://github.com/code-dot-org/code-dot-org/pull/22307/files#diff-3c5e6748726d3411afe8256d6ac20a3eR463 for how I intend to use these new options.

### DIFF
--- a/core/ui/function_editor.js
+++ b/core/ui/function_editor.js
@@ -83,6 +83,8 @@ Blockly.FunctionEditor = function(
   }
   this.parameterBlockTypes = opt_parameterBlockTypes || {};
   this.disableParamEditing = opt_disableParamEditing || false;
+
+  Blockly.FunctionEditor.allFunctionEditors.push(this);
 };
 
 Blockly.FunctionEditor.BLOCK_LAYOUT_LEFT_MARGIN = Blockly.BlockSpaceEditor.BUMP_PADDING_LEFT;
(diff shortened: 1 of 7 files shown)


joshlory:
Worth adding a URL tweak to playground.html for showing the Behavior Editor?

core/ui/function_editor.js
```diff
@@ -96,6 +109,20 @@ Blockly.FunctionEditor.RTL_CLOSE_BUTTON_OFFSET = 5;
...
+ * Get a string from Blockly.Msg, unless overridden by msgOverrides
+ */
+Blockly.FunctionEditor.prototype.getMsg = function(label) {
+  return this.msgOverrides_[label] || Blockly.Msg[label];
+};
+
+Blockly.FunctionEditor.prototype.paramEditingDisabled = function () {
+  return Blockly.disableParamEditing || this.disableParamEditing;
+};
+
```
joshlory: Nit: this maybe could've been extracted to a separate PR.

balderdash:
The changes in this PR are all optional customizations. Just try to check that I'm not changing any default behavior and that should be good enough.

PR #106, opened 2018-05-11
Cross-tab copy / paste

joshlory:
copy

### DIFF
--- a/core/ui/block_space/block_space.js
+++ b/core/ui/block_space/block_space.js
@@ -702,8 +702,8 @@ Blockly.BlockSpace.prototype.fireChangeEvent = function() {
  */
 Blockly.BlockSpace.prototype.paste = function(clipboard) {
   var xmlBlock = clipboard.dom;
-  // When pasting into a different block spaces, remove parameter blocks
-  if (this !== clipboard.sourceBlockSpace) {
+  // When pasting into the main block space, remove parameter blocks
+  if (this === Blockly.mainBlockSpace) {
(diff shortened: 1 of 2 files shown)


Hamms:
I thought we had reasons for not wanting to do this; preventing students from pasting blocks into levels that don't have those blocks in the toolbox, etc

balderdash:
Josh added better handling for unknown block types, so pasting unsupported blocks will just give you an ugly gray "unknown block" block rather than completely breaking blockly.

The curriculum question is another matter. We could make it a level option to enable/disable pasting, but I'm not sure if it would be even more annoying to have copy and paste that only works *sometimes*.

Hamms:
@ryansloan @kiki-lee can you remind me of the potential issues with allowing this kind of copy-pasting?

kiki-lee:
I think the biggest concern was cheating, but also I've been told that this is just really hard to do.

kiki-lee:
I suppose someone could copy blocks from one skin to another (where they're not supported) and throw themselves into an error...but we'll need to figure that out before we move to a backpack, anyhow :)

ryansloan:
I think that it's ok to ship this as-is, given that most of our K-5 students haven't quite figured out all this keyboard copy/paste business, and _most_ of the time having this ability wouldn't be a big issue anyway. We should address the cheating issue as part of the backpack work that we do in a few months..

kiki-lee:
Quick FYI that I don't worry at all about the cheating aspect.  I think the amount that it requires students to think about other ways of solving the problem far outweighs the risk of them not having to manually enter expected code.

PR #108, opened 2018-05-15
Ignore duplicate touch events

balderdash:
There's some special touch event handling code here that makes the toolbox more responsive than simply relying on simulated mouse events. Unfortunately chrome is giving us two touch events (so it closes the toolbox category immediately after you open it), so I'm adding a check to ignore duplicated touch events.

### DIFF
--- a/core/ui/block_space/toolbox.js
+++ b/core/ui/block_space/toolbox.js
@@ -269,15 +269,7 @@ Blockly.Toolbox.TreeControl.prototype.enterDocument = function() {
       'onpointerdown' in window ||
       'onmspointerdown' in window) {
     var el = this.getElement();
-    var debouncedHandler =
-      goog.functions.debounce(this.handleTouchEvent_, 50, this);
-    var handler = (function (e) {
-      if (this.getNodeFromEvent_(e)) {
-        e.preventDefault();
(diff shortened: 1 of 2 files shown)


balderdash:
It's a little weird, because:
* we still need to call `e.preventDefault()` on each call or the browser will fire a mouse event too
* AFAICT there's no way to add a debounced method to the `TreeControl` prototype and ensure that `this` remains properly bound inside.

Any ideas on how this could be cleaned up?

Hamms:
Ah dang, I didn't think about `preventDefault`. Does using `goog.functions.throttle` rather than debounce make that easier?

balderdash:
I'm afraid not, since `throttle` will just delay rather than drop the second handler call.

`rateLimit` does what we want though, but I had to upgrade our version of `google-closure-library` a bit to use it. The latest version of the library gives us a bunch of build errors, so I only upgraded to the release that added `rateLimit` :disappointed:

Hamms:
Oh, awesome! Yeah, `rateLimit` seems to be what I thought `throttle` was. Shame we're still stuck on a version of closure that's over a year old, but definitely a step in the right direction.

The final version of this looks great!

PR #109, opened 2018-05-17
The 'STACK' of a functional_definition is the implicit return value

joshlory:
Regular `procedures_defreturn` functions have an explicit return connection:

https://github.com/code-dot-org/blockly/blob/d18d42d5ed6404dc64e8f461650245c2c6a91f47/blocks/procedures.js#L301

image

But for `functional_definition` the 'STACK' connection is always the implicit return value.

Fixes https://github.com/code-dot-org/code-dot-org/pull/22463#issuecomment-390008959.


joshlory:
I don't think we could be (famous last words...) since the block doesn't have a 'RETURN' input connection.

PR #114, opened 2018-05-21
Use output type to set default block color, when no color is specified

joshlory:
One thing I found during this process:

The only place we use multiple type checks (vs. None or exactly 1 type) is for the the `text_length` / `list_length` and `text_empty` / `list_empty` blocks.  https://github.com/code-dot-org/blockly/blob/261ac3f64b7d5f0bb3303c6104d0e954009c2f34/blocks/lists.js#L179

Since they're separate blocks, we could collapse them to only take the corresponding type.  I think it's just an ease-of-use feature to allow them to be used interchangeably on strings and lists.

### DIFF
--- a/core/connections/connection.js
+++ b/core/connections/connection.js
@@ -432,10 +432,14 @@ Blockly.Connection.prototype.colorForType = function(checks) {
       return [160, 0.45, 0.65];
     case 'Number':
       return [258, 0.35, 0.62];
+    case 'Boolean':
+      return [196, 1.0, 0.79];
+    case 'Array':
+      return [40, 1.0, 0.99];
     case 'Colour':
       return [196, 1.0, 0.79];
     case 'Sprite':
-      return [0, 0.45, 0.65];
+      return [355, 0.70, 0.70];
     case 'Behavior':
       return [136, 0.84, 0.80];
     case 'Location':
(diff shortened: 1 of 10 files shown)


core/connections/connection.js
```diff
@@ -772,6 +776,12 @@ Blockly.Connection.prototype.setCheck = function(check, opt_strict) {
       // Bump away.
       this.sourceBlock_.bumpNeighbours_();
     }
+
+    // If the setting an output check, and the block has no color, make the
+    // block the default color for the given output type.
+    if (this.type === Blockly.OUTPUT_VALUE && !this.sourceBlock_.getColour()) {
+      this.sourceBlock_.setHSV.apply(this.sourceBlock_, this.colorForType(check));
```
Hamms: Why is this happening here and not in `setOutput`?

joshlory: I had it there originally (see commit 318d78f0e2ddefa6a1de1c836e32472251d05e49) but there are some blocks that set different checks based on block state:

https://github.com/code-dot-org/blockly/blob/04705fd7494610e1f6a2b682b3f33ebbc3fb3161/blocks/text.js#L460-L464

My thought was that we might want to change block color based on dynamic outputs.  What do you think is better?

joshlory: (Right now it doesn't make a difference — once the block has a default color, changing the output check won't overwrite that color with a new one.)

PR #115, opened 2018-05-22
Gracefully handle code generation for unknown blocks

joshlory:
Follow-up to PR https://github.com/code-dot-org/blockly/pull/101.  Sample code generation:

image

```js
for (var count = 0; count < 10; count++) {
  /* Unknown block: draw_turn_by_constant_dropdown */
}
```

### DIFF
--- a/core/code_generation/generator.js
+++ b/core/code_generation/generator.js
@@ -220,8 +220,7 @@ Blockly.CodeGenerator.prototype.blockToCode = function(block, opt_showHidden) {
 
   var func = this[block.type];
   if (!func) {
-    throw 'Language "' + this.name_ + '" does not know how to generate code ' +
-        'for block type "' + block.type + '".';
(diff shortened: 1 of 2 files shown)


core/code_generation/generator.js
```diff
@@ -220,8 +220,7 @@ Blockly.CodeGenerator.prototype.blockToCode = function(block, opt_showHidden) {
 
   var func = this[block.type];
   if (!func) {
-    throw 'Language "' + this.name_ + '" does not know how to generate code ' +
-        'for block type "' + block.type + '".';
+    return this.comment('Unknown block: ' + block.type);
```
balderdash: Worth firing another `unknownBlock` event here? It's unlikely, but if somebody adds a block and forgets the corresponding generator, we could hit this case without first hitting the unknown block initialization code.

PR #118, opened 2018-06-01
Typed params in the Behavior Editor

joshlory:
Only create different blocks for "Sprite" and "Location".  For now the other options all create the same untyped `parameter_get` block.

image

### DIFF
--- a/blocks/variables.js
+++ b/blocks/variables.js
@@ -126,7 +126,8 @@ Blockly.Blocks.parameters_get = {
     if (type) {
       // Clear current color, so we can infer it from output type.
       this.setColour(null);
-      this.outputConnection.setCheck(type, true);
+      var strict = !!Blockly.valueTypeTabShapeMap[type];
+      this.outputConnection.setCheck(type, strict);
     }
   },
   renameVar: function(oldName, newName) {
--- a/blocks/variables.js
+++ b/blocks/variables.js
@@ -104,6 +104,7 @@ Blockly.Blocks.parameters_get = {
     // Must be marked EDITABLE so that cloned blocks share the same var name
     fieldLabel.EDITABLE = true;
     this.setHelpUrl(Blockly.Msg.VARIABLES_GET_HELPURL);
+    this.setHSV(7, 0.80, 0.95);
     this.appendDummyInput()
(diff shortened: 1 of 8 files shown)


joshlory:
PR updated: also mutate normal params based on their specified type.

image

blocks/variables.js
```diff
@@ -113,6 +113,22 @@ Blockly.Blocks.parameters_get = {
...
+      container.setAttribute('output', checks[0]);
+      return container;
+    }
+  },
+  domToMutation: function(container) {
+    var type = container.getAttribute('output');
+    if (type) {
+      // Clear current color, so we can infer it from output type.
+      this.setColour(null);
+      this.outputConnection.setCheck(type, true);
```
balderdash: This check shouldn't be strict for numbers/booleans/strings.

However, it *should* be strict for sprites and locations, and I think that would eliminate our need for type-specific param blocks (except maybe the `this sprite` block, which is a little different).

joshlory:
@balderdash ok I think this is ready for review!

core/ui/function_editor.js
```diff
@@ -374,7 +381,8 @@ Blockly.FunctionEditor.prototype.paramsAsParallelArrays_ = function() {
     paramNames.push(blockXML.firstElementChild.textContent);
     paramIDs.push(paramID);
     if (blockXML.childNodes.length > 1) {
-      paramTypes.push(blockXML.childNodes[1].textContent);
+      var node = blockXML.childNodes[1];
+      paramTypes.push(node.getAttribute('output') || node.textContent);
```
balderdash: The output attribute holds the new param types, while textContent is used for CSinA types, right? Worth adding a comment to explain this bit of weirdness?

tests/playground.html
```diff
@@ -126,8 +126,19 @@
       FUNCTION_DESCRIPTION_LABEL: 'What kind of ocean is this supposed to boil?',
     },
     'ocean_boiler_definition',
-    null,
-    true,
+    {
+      [Blockly.BlockValueType.SPRITE]: 'sprite_variables_get',
+      [Blockly.BlockValueType.LOCATION]: 'location_variables_get'
```
balderdash: Since the parameter block can now set its output type, do we still need to specify separate param blocks for the strict types?

PR #120, opened 2018-06-14
When pressing ctrl/cmd, only drag a single block (not the full stack)

joshlory:
extract-block

Trying out an idea from https://docs.google.com/document/d/1wJMKXJS6Z8tsYQRRq93r7k4TTzG4jQbIXfKdKcSTZTI/edit


joshlory:
Eventually, yes!

joshlory:
Ah, interesting:  Scratch 3.0 is looking at something similar, using right click instead of cmd/ctrl + click.

https://github.com/LLK/scratch-gui/issues/1708

I like their idea of long-press on mobile to extract one block, although novice users might do this by accident and get confused.  Our story might be to instead implement multi-touch block separation for the mobile scenario.

PR #125, opened 2018-06-20
Allow pasting blocks from other applications

joshlory:
For example, pasting raw XML from Slack into a block space.  A user should be able to copy the following to their clipboard, then paste it into Blockly:

```xml
<xml>
  <block type="colour_blend" inline="false"></block>
</xml>
```

1. Is there XML on the clipboard?  If yes, use that.
2. Otherwise use the plain text from the clipboard.

https://docs.google.com/document/d/1wJMKXJS6Z8tsYQRRq93r7k4TTzG4jQbIXfKdKcSTZTI/edit#heading=h.njzku5ndadkc

### DIFF
--- a/core/ui/block_space/block_space_editor.js
+++ b/core/ui/block_space/block_space_editor.js
@@ -761,7 +761,7 @@ Blockly.BlockSpaceEditor.prototype.onCutCopy_ = function(e) {
  */
 Blockly.BlockSpaceEditor.prototype.onPaste_ = function(e) {
   try {
-    var xml = Blockly.Xml.textToDom(e.clipboardData.getData('text/xml'));
+    var xml = Blockly.Xml.textToDom(e.clipboardData.getData('text/xml') || e.clipboardData.getData('text/plain'));
     Blockly.focusedBlockSpace.paste({
       dom: xml.firstElementChild
     });


joshlory:
Right now this only pastes the first block in the XML snippet.  I want to update it to paste multiple unconnected blocks from the clipboard (for example, the result of serializing blocks in another block space).

core/ui/block_space/block_space.js
```diff
@@ -698,57 +698,25 @@ Blockly.BlockSpace.prototype.fireChangeEvent = function() {
...
   }
-  var block = Blockly.Xml.domToBlock(this, xmlBlock);
-  // Move the duplicate to original position.
-  var blockX = parseInt(xmlBlock.getAttribute('x'), 10);
-  var blockY = parseInt(xmlBlock.getAttribute('y'), 10);
-  if (!isNaN(blockX) && !isNaN(blockY)) {
-    if (Blockly.RTL) {
-      blockX = -blockX;
-    }
-    // Offset block until not clobbering another block.
```
joshlory: This is now handled by `bumpNeighbours_`, but depends on the changes in PR https://github.com/code-dot-org/blockly/pull/128.

core/ui/block_space/block_space.js
```diff
@@ -698,57 +698,25 @@ Blockly.BlockSpace.prototype.fireChangeEvent = function() {
...
-      }
-    } while (collide);
-    block.moveBy(blockX, blockY);
-  }
-  block.setUserVisible(true);
-  block.select();
+  var blocks = Blockly.Xml.domToBlockSpace(this, xml);
+
+  blocks.forEach(function(b) {
+    b.blockly_block.bumpNeighbours_();
```
Hamms: nit: `bumpNeighbours_` should probably be renamed to no longer be pseudoprivate

joshlory: Fixed in commit dd076dec045e0a5c1248d5fc65a06318f567497c 👍

PR #127, opened 2018-06-21
Stop reading & serializing `inline` attribute for block inputs

joshlory:
In almost every case we want the block default for "inline", not what the user has saved in the Level Sources table.

### DIFF
--- a/core/utils/xml.js
+++ b/core/utils/xml.js
@@ -97,7 +97,6 @@ Blockly.Xml.blockToDom = function(block, ignoreChildBlocks) {
     element.appendChild(commentElement);
   }
 
-  var setInlineAttribute = false;
   for (i = 0; i < block.inputList.length; i++) {
     input = block.inputList[i];
     var empty = true;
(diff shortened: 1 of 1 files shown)


Hamms:
I believe this will also ignore levelbuilder-defined "inline" settings, yes? Presumably we want to do strictly for level sources.

joshlory:
In a world where block definitions are editable by level builders, I think we just want one place to configure that setting (and have it apply to all blocks).

This change would always respect what's in the block definition, no longer let it be overridden by the XML attribute.  If we think a block works best a certain way, it should always render that way.

Hamms:
Is that actually what we want, though? I would definitely expect that we _usually_ want a block to be rendered one way or the other, but might want to situationally override that for certain levels.

I'd at the very least want to check with @kiki-lee to confirm that is not a tool she uses before we take it out of the toolbox.

kiki-lee:
I'm good with this.  I didn't ever use the xml attribute to change the block.
It seems like blocks that are too long need to be shortened all over the place, one level is not likely going to have space that another doesn't.

joshlory:
We can clean up levelbuilder toolboxes, but I think it has diminishing value... there will also be millions of levelsource rows that use the deprecated attribute.

PR #128, opened 2018-06-22
Disambiguate (bump) connections of the same type

joshlory:
Until now, `bumpNeighbours_` has only looked at opposing connections.  This PR updates the definition of "neighbors" to include situations where two blocks have an overlapping connection of the _same_ type.

In this case, since neither connection is "superior", we add new logic to determine which to bump:
1. Prioritize bumping unconnected blocks over connected blocks. 
bump-priority-unconnected
bump-priority-connected

2. When both blocks are unconnected, prioritize bumping blocks lower on the screen.
bump-priority-position

### DIFF
--- a/core/connections/connection.js
+++ b/core/connections/connection.js
@@ -835,19 +835,20 @@ Blockly.Connection.prototype.neighbours_ = function(maxLimit) {
 
   // Construct a new connection DB, with matching and opposing connections.
   var db = [];
-  var a = 0, b = 0;
+  var a = 0, b = 0, connection;
   while (a < db1.length || b < db2.length) {
     if (!db2[b] || (db1[a] && db1[a].y_ < db2[b].y_)) {
-      db.push(db1[a]);
+      connection = db1[a];
+      if (connection !== this) {
+        db.push(connection)
+      }
       a++;
     } else {
-      db.push(db2[b]);
+      connection = db2[b];
+      if (connection !== this) {
+        db.push(connection)
+      }
       b++;
-    }
-  }
-  for (var i = 0; i < db.length; i++) {
-    if (i === this) {
-      db.splice(i, 1);
     }
   }
 
--- a/core/connections/connection.js
+++ b/core/connections/connection.js
@@ -393,6 +393,7 @@ Blockly.Connection.prototype.bumpAwayFrom_ = function(staticConnection) {
     dx = -dx;
   }
   rootBlock.moveBy(dx, dy);
+  rootBlock.bumpNeighbours_();
 };
 
 /**
--- a/core/connections/connection.js
+++ b/core/connections/connection.js
@@ -829,7 +829,26 @@ Blockly.Connection.prototype.getCheck = function () {
 Blockly.Connection.prototype.neighbours_ = function(maxLimit) {
   // Determine the opposite type of connection.
   var oppositeType = Blockly.OPPOSITE_TYPE[this.type];
-  var db = this.dbList_[oppositeType];
+  var db1 = this.dbList_[this.type];
+  var db2 = this.dbList_[oppositeType];
+
+  // Construct a new connection DB, with matching and opposing connections.
+  var db = [];
+  var a = 0, b = 0;
+  while (a < db1.length || b < db2.length) {
+    if (!db2[b] || (db1[a] && db1[a].y_ < db2[b].y_)) {
+      db.push(db1[a]);
+      a++;
+    } else {
+      db.push(db2[b]);
+      b++;
+    }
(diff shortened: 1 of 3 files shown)


joshlory:
Old bump behavior, for comparison:

old-bump-unconnected

old-bump-connected

old-bump-position

core/ui/block.js
```diff
@@ -1357,15 +1357,35 @@ Blockly.Block.prototype.bumpNeighbours_ = function() {
...
+      } else if (!connection.targetConnection && otherConnection.targetConnection) {
+        connection.bumpAwayFrom_(otherConnection);
+      // Otherwise bump the block that is lower on the screen.
+      } else {
+        var rootY = rootBlock.getRelativeToSurfaceXY().y;
+        var otherY = otherRootBlock.getRelativeToSurfaceXY().y;
+        if (rootY > otherY) {
+          connection.bumpAwayFrom_(otherConnection);
+        } else {
+          otherConnection.bumpAwayFrom_(connection);
```
Hamms: nit: it might be slightly clearer to do something like

```javascript
var bumpOther;

// When bumping connections of opposite types, always bump the inferior block.
if (connection.type !== otherConnection.type) {
  bumpOther = connection.isSuperior();
// If one is connected and the other is unconnected, always bump the
// unconnected block.
} else if (connection.targetConnection && !otherConnection.targetConnection) {
  bumpOther = true;
} else if (!connection.targetConnection && otherConnection.targetConnection) {
  bumpOther = false;
// Otherwise bump the block that is lower on the screen.
} else {
  var rootY = rootBlock.getRelativeToSurfaceXY().y;
  var otherY = otherRootBlock.getRelativeToSurfaceXY().y;
  bumpOther = otherY >= rootY;
}

if (bumpOther) {
  otherConnection.bumpAwayFrom_(connection);
} else {
  connection.bumpAwayFrom_(otherConnection);
}
```

core/connections/connection.js
```diff
@@ -829,7 +830,26 @@ Blockly.Connection.prototype.getCheck = function () {
 Blockly.Connection.prototype.neighbours_ = function(maxLimit) {
   // Determine the opposite type of connection.
   var oppositeType = Blockly.OPPOSITE_TYPE[this.type];
-  var db = this.dbList_[oppositeType];
+  var db1 = this.dbList_[this.type];
+  var db2 = this.dbList_[oppositeType];
+
+  // Construct a new connection DB, with matching and opposing connections.
```
Hamms: Given how aggressively optimized the existing connection DB operations are, I'm not sure I love how much work this has to do to build out the new connection DB. Have you been able to do any kind of performance analysis to confirm that it's fine to add a linear operation to an otherwise (I believe) entirely logarithmic set of operations?

joshlory: Great question.  I this case I'm doing a merge of two already-sorted lists, so I don't think the added linear operation should be too bad?  It's certainly better than concatenating the two arrays together and sorting them.

Hamms: yeah, it's definitely not too bad. I only even bring it up because everything else in this area is _so_ optimized that I worry it's optimized for a good reason

core/ui/block.js
joshlory: Good call, cleaned up in db255ba1996aa19083115f8eca66a71322515afb!  I intentionally changed `otherY >= rootY` -> `otherY > rootY` to account for the fix I want to eventually apply in https://github.com/code-dot-org/blockly/pull/125: pasted blocks should be bumped, not the block the land on top of.

core/connections/connection.js
```diff
@@ -829,7 +830,26 @@ Blockly.Connection.prototype.getCheck = function () {
...
+    if (!db2[b] || (db1[a] && db1[a].y_ < db2[b].y_)) {
+      db.push(db1[a]);
+      a++;
+    } else {
+      db.push(db2[b]);
+      b++;
+    }
+  }
+  for (var i = 0; i < db.length; i++) {
+    if (i === this) {
```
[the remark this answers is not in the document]
joshlory: Fixed in db255ba1996aa19083115f8eca66a71322515afb.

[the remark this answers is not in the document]
joshlory: (The reason this wasn't causing problems before is that we ignore connections from the same root tree.  The root blocks will be the same if it's the same block 😄.  But this is still good to fix, so we don't add extraneous connections to the returned neighbors.)

https://github.com/code-dot-org/blockly/blob/5539c8cb16d170a44c4c0bd124a6b1940cd78ef6/core/ui/block.js#L1361-L1362

PR #129, opened 2018-06-28
Don't fire unnecessary window "resize" events

joshlory:
Reduces the time to open the "Sprites" tab of the toolbox on /p/spritelab from ~500ms to ~200ms.

**Before:**

screen shot 2018-06-27 at 1 57 55 pm

**After:**

screen shot 2018-06-27 at 5 12 24 pm

Both graphs have a horizontal time scale of about ~570ms.

### DIFF
--- a/core/utils/xml.js
+++ b/core/utils/xml.js
@@ -525,10 +525,10 @@ Blockly.Xml.domToBlock = function(blockSpace, xmlBlock) {
   var next = block.nextConnection && block.nextConnection.targetBlock();
   if (next) {
     // Next block in a stack needs to square off its corners.
-    // Rendering a child will render its parent.
-    next.render();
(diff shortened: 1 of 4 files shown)


core/utils/xml.js
```diff
@@ -525,10 +525,10 @@ Blockly.Xml.domToBlock = function(blockSpace, xmlBlock) {
...
   if (next) {
     // Next block in a stack needs to square off its corners.
-    // Rendering a child will render its parent.
-    next.render();
+    next.render(true);
+    block.render(true);
   } else {
-    block.render();
+    block.render(true);
   }
```
joshlory: If we were to optimize further, we could look into not rendering each `domToBlock` call — or batch renders when possible.

PR #131, opened 2018-06-29
Remove `Blockly.Comment` namespace and bubble comment impl

joshlory:
The Education team is choosing to implement comments as standalone blocks.

image


joshlory:
This removes the old style of embedded comment:

image

PR #133, opened 2018-06-30
Wait to render "unused block" DOM until it's needed

joshlory:
This is a slightly different approach to fixing the issue in PR https://github.com/code-dot-org/blockly/pull/11, to unblock PR https://github.com/code-dot-org/blockly/pull/129.

By not rendering the block in `this.setIsUnused();` we're able to postpone rendering until the full stack of blocks is built up, which will avoid any potential unwanted bumps.

### DIFF
--- a/core/ui/block_space/block_space.js
+++ b/core/ui/block_space/block_space.js
@@ -350,7 +350,7 @@ Blockly.BlockSpace.prototype.createDom = function() {
   Blockly.bindEvent_(this.svgBlockCanvas_, Blockly.BlockSpace.EVENTS.RUN_BUTTON_CLICKED, this, function () {
     this.getTopBlocks().forEach(function (block) {
       if (block.isUnused()) {
-        block.svg_.addUnusedFrame();
+        block.getSvgRenderer().addUnusedFrame();
       }
     });
   });
--- a/core/ui/block_space/block_space.js
+++ b/core/ui/block_space/block_space.js
@@ -347,6 +347,14 @@ Blockly.BlockSpace.prototype.createDom = function() {
   }
   this.svgDragCanvas_ = Blockly.createSvgElement('g', {'class': 'svgDragCanvas'}, Blockly.dragSvg);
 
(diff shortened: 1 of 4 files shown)


core/ui/block_svg/block_svg_unused.js
```diff
@@ -143,6 +134,10 @@ Blockly.BlockSvgUnused.prototype.render = function (svgGroup) {
   goog.dom.insertChildAt(svgGroup, this.frameGroup_, 0);
 
   Blockly.addClass_(this.frameGroup_, 'hidden');
+  var frameGroup = this.frameGroup_;
+  setTimeout(function () {
+    Blockly.removeClass_(frameGroup, 'hidden');
+  }, 0);
```
Hamms: what up with this?

core/ui/block_space/block_space.js
```diff
@@ -347,6 +347,14 @@ Blockly.BlockSpace.prototype.createDom = function() {
   }
   this.svgDragCanvas_ = Blockly.createSvgElement('g', {'class': 'svgDragCanvas'}, Blockly.dragSvg);
 
+  Blockly.bindEvent_(this.svgBlockCanvas_, Blockly.BlockSpace.EVENTS.RUN_BUTTON_CLICKED, this, function () {
+    this.getTopBlocks().forEach(function (block) {
+      if (block.isUnused()) {
+        block.svg_.addUnusedFrame();
```
Hamms: nit: this should be `block.getSvgRenderer().addUnusedFrame()`

core/ui/block_svg/block_svg_unused.js
joshlory: Unfortunately CSS `@keyframes` aren't well-supported cross browser yet 😕.  This triggers the CSS transition to fade in opacity:

https://github.com/code-dot-org/blockly/blob/dd076dec045e0a5c1248d5fc65a06318f567497c/core/ui/css.js#L145-L148

Hamms: Ooooooh yeah. I think that might be why I set this up to be initialized with a hidden frame in the first place. Mind adding a comment explaining what's going on here?

PR #135, opened 2018-07-09
Run closure compiler under webpack

balderdash:


webpack.config.js
```diff
@@ -0,0 +1,34 @@
...
+  output: {
+    path: path.resolve(__dirname, 'dist'),
+    filename: 'blockly.js',
+  },
+  plugins: [
+    new ClosureCompilerPlugin({
+      compiler: {
+        jar: 'node_modules/google-closure-compiler/compiler.jar',
+        compilation_level: 'SIMPLE',
+        entry_point: 'BlocklyModule',
```
Hamms: can you explain why we need `BlocklyModule` and can't just have the entry point be `Blockly`?

webpack.config.js
```diff
@@ -0,0 +1,34 @@
...
+
+module.exports = {
+  name: 'blockly',
+  mode: 'development',
+  devtool: 'source-map',
+  entry: {
+    blockly: path.join(__dirname, 'core/module.js'),
+  },
+  output: {
+    path: path.resolve(__dirname, 'dist'),
```
Hamms: is there a specific reason we we `join` for `entry` and `resolve` for `output`?

tests/playground.html
```diff
@@ -4,6 +4,7 @@
 <meta charset="utf-8">
 <title>Blockly Playground</title>
 <script type="text/javascript" src="../node_modules/google-closure-library/closure/goog/base.js"></script>
+<!--<script type="text/javascript" src="./test_dependency_map.js"></script>-->
```
Hamms: nit: looks like a dev remnant

webpack.config.js
```diff
@@ -0,0 +1,34 @@
+const path = require('path');
+const ClosureCompilerPlugin = require('webpack-closure-compiler');
+
+module.exports = {
+  name: 'blockly',
+  mode: 'development',
```
Hamms: do we really want to default to dev and not prod?

webpack.config.js
balderdash: The playground directly loads the rest of the uncompiled source files, but you can't use a JS file in the browser if it includes a module declaration, so I pulled that out into a separate file.

webpack.config.js
balderdash: Fixed.

webpack.config.js
balderdash: Nope.

package.json
```diff
@@ -11,21 +12,24 @@
...
     "i18n",
     "media"
   ],
   "scripts": {
-    "build": "./deploy.sh",
-    "build:dev": "./deploy.sh debug",
+    "build": "webpack-cli",
+    "build:dev": "DEV=1 webpack-cli",
     "preversion": "npm run test",
     "version": "npm run build:dev && npm run build && git add -A build-output",
```
Hamms: nit: can remove the `git add` step here

tests/playground.html
```diff
@@ -5,6 +5,7 @@
 <title>Blockly Playground</title>
 <script type="text/javascript" src="../node_modules/google-closure-library/closure/goog/base.js"></script>
 <script type="text/javascript" src="./test_dependency_map.js"></script>
+<script type="text/javascript" src="./test_dependency_map.js"></script>
```
Hamms: looks like this is being included twice now? I think the last change should have been to remove this line rather than uncomment it

PR #137, opened 2018-07-12
Allow un-deletable blocks to be highlighted

joshlory:
Found on level https://studio.code.org/s/course3/stage/14/puzzle/3 via PR https://github.com/code-dot-org/code-dot-org/pull/23677.

We don't use un-deletable blocks in the new courses AFAIK, so I don't think shipping this is blocking the merge of the linked PR.


joshlory:
A bit more conversation here: https://codedotorg.slack.com/archives/C0T0GC9GE/p1531437352000287

PR #138, opened 2018-07-17
Stop (indirectly) using goog.global.setTimeout

balderdash:
By default, `goog.global` gets set to `window`, but with my earlier build changes that is no longer the case. There are a few references to `goog.global.setTimeout` in the closure library that break as a result (`goog.Timer.callOnce` and `goog.function.rateLimit` both use it), so I'm just removing uses of those library methods.

`goog.async.AnimationDelay` grabs the `window` object directly instead of using `goog.global`, so it still works.

Fun fact, blockly made basically the same change upstream: https://github.com/google/blockly/issues/1905


core/ui/block_space/toolbox.js
```diff
@@ -286,6 +287,16 @@ Blockly.Toolbox.TreeControl.prototype.handleTouchEvent_ = function(e) {
...
                e.type === goog.events.EventType.MSPOINTERDOWN)) {
+
+    // Rate limit to once every 50ms
+    if (this.touchRateLimited) {
+      return;
+    }
+    this.touchRateLimited = true;
+    setTimeout(function () {
+      this.touchRateLimited = false;
+    }.bind(this));
```
Hamms: I'm a little worried about the rate limiting happening _after_ the prevent default; can you confirm whether or not that has any impact on functionality?

balderdash: It *seems* to work fine, but I might as well move the rate limiting further up.

PR #140, opened 2018-07-18
Still load node_modules cache on package.json mismatch

balderdash:
Circle builds have been slower than normal because of all the package.json changes. If it changes, it still speeds things up if we load the most recent node_modules cache.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,7 @@ jobs:
     - restore_cache:
         keys:
         - v3-dep-{{ checksum "package.json" }}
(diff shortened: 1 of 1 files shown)


Hamms:
doesn't this have the potential to break things if we load an old cache?

balderdash:
Hmm, I assumed that the subsequent `npm install` would set things right if the cache was out of date, but it looks like won't actually remove outdated packages.

I suppose an extra minute isn't too high a price to pay to avoid weird issues arising from that.

PR #143, opened 2018-08-09
Handle the default case in `allProcedures()` for other proc types

joshlory:
Better fix for PR https://github.com/code-dot-org/code-dot-org/pull/24177.


balderdash:
I looked into this a little more, because this looks like the right thing to do, but obviously it isn't how behavior categories currently get auto-populated.

The flyout creation code falls through to this case when the category has the attribute `custom="Behavior"`:
https://github.com/code-dot-org/blockly/blob/85815822fb444e8eba4ba5b88b057551595ccdbe/core/ui/block_space/flyout.js#L491-L504

This calls `Blockly.Variables.flyoutCategory`, which finds all the behavior variables, and creates corresponding getter blocks (based on the setter block indicated elsewhere by `Blockly.Variables.registerGetter`).

The method you're modifying is used in `Blockly.Procedures.flyoutCategory`, which is used in a previous case in the flyout creation code, here:
https://github.com/code-dot-org/blockly/blob/85815822fb444e8eba4ba5b88b057551595ccdbe/core/ui/block_space/flyout.js#L460-L481

However, it isn't go into that case because the category custom attribute is "Behavior", not "PROCEDURE". That is, unless `Blockly.topLevelAutoPopulate` is set, which is the case for Caldwell's level:
image

That setting is why the behaviors weren't showing up, and it's also messing up the rest of the toolbox by turning every category into a function category:
function_categories

So a few conclusions:
* I turned off that `topLevelAutoPopulate` setting for the level. That brings back the shared behaviors, but the getter blocks still don't accept the additional param
* Restrict the `topLevelAutoPopulate` setting so that it can only take effect if the toolbox actually has no categories
* We still need something along the lines of this PR in order to correctly create the behavior getter blocks with params

balderdash:
Sorry for the very long comment :sweat_smile:

joshlory:
Thank you for the detailed write-up!  I was missing quite a few things in the original PR, which should be addressed now.

It feels semi-gross to add this behavior-specific logic to the general Blockly repo.  But on the flip side, I'm not sure we want to expose the number of hooks we'd need to configure this behavior outside of Blockly, just for this one case.  Curious to hear your thoughts!

balderdash:
The flyout initialization code accepts an arbitrary `initialize` function, which is currently used to add the "Create a Behavior" button in the Behavior category:
https://github.com/code-dot-org/code-dot-org/blob/92996dab165f2f6e196ee70650b8147d4e43070d/apps/src/gamelab/blocks.js#L339-L347

You could add a config option to _only_ run the `initialize` method (skipping the the default autopopulation code), and then move the behavior-specific code from this PR to that initialize method.

If you don't go that route, you should probably just get rid of that config option (it's only used for behaviors) and move the button adding code to the behavior-specific code here.

PR #145, opened 2018-08-24
Add editor methods to lock/unlock block movement

balderdash:
This adds lock/unlock movement methods to the block space editor to support limited autorun in https://github.com/code-dot-org/code-dot-org/pull/24196.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -5,7 +5,7 @@ const DEV = process.env.DEV;
 
 const compilerOptions = {
   jar: 'node_modules/google-closure-compiler/compiler.jar',
-  compilation_level: DEV ? 'WHITESPACE_ONLY' : 'SIMPLE',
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -5,7 +5,7 @@ const DEV = process.env.DEV;
 
 const compilerOptions = {
   jar: 'node_modules/google-closure-compiler/compiler.jar',
-  compilation_level: DEV ? 'WHITESPACE_ONLY' : 'SIMPLE',
+  compilation_level: 'SIMPLE',
```
Hamms: what's this change for?

balderdash: Right, this is unrelated. Compiling in `WHITESPACE_ONLY` mode wasn't working, and since `SIMPLE` mode doesn't rename variables it's basically just as good for development.

PR #147, opened 2018-09-07
Update infinite loop error to an Error object

balderdash:
For consistency, I'm updating this to match the error thrown for infinite loops here: https://github.com/code-dot-org/code-dot-org/blob/7e2fcc28fbb791f0dd594c8b6e6192bca6f9a277/apps/src/lib/tools/jsinterpreter/CustomMarshalingInterpreter.js#L11

### DIFF
--- a/generators/javascript/loops.js
+++ b/generators/javascript/loops.js
@@ -117,7 +117,7 @@ Blockly.JavaScript.controls_for = function() {
       code += (up ? ' += ' : ' -= ') + step;
       if (step === 0) {
         // guaranteed to loop infinitely if we're looping with step
-        branch = '  throw Infinity;\n' + branch;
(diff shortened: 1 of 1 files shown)


generators/javascript/loops.js
```diff
@@ -117,7 +117,7 @@ Blockly.JavaScript.controls_for = function() {
       code += (up ? ' += ' : ' -= ') + step;
       if (step === 0) {
         // guaranteed to loop infinitely if we're looping with step
-        branch = '  throw Infinity;\n' + branch;
+        branch = '  throw new Error("Infinity");\n' + branch;
```
joshlory: Wonder if this should be provided by the consumer, so it doesn't fall out of sync with https://github.com/code-dot-org/code-dot-org/pull/24634/files#r216867192 again?

Probably not a big deal either way...

PR #156, opened 2018-11-15
Defend against empty variable names

joshlory:
With guidance from @Bjvanminnen, alternate fix for PR https://github.com/code-dot-org/code-dot-org/pull/26085.

### DIFF
--- a/tests/variables_test.js
+++ b/tests/variables_test.js
@@ -65,3 +65,17 @@ function test_allVariablesOtherCategory() {
 
   goog.dom.removeNode(container);
 }
+
+function test_cannotRenameVariablesToEmpty() {
+  var container = Blockly.Test.initializeBlockSpaceEditor();
+
+  var blockXML = '<xml><block type="variables_get"><title name="VAR">i</title></block></xml>';
+  Blockly.Xml.domToBlockSpace(Blockly.mainBlockSpace, Blockly.Xml.textToDom(blockXML));
+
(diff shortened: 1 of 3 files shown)


Bjvanminnen:
With this change here https://github.com/code-dot-org/blockly/blob/master/core/ui/fields/field_variable.js#L156 we will still call `setText` even if `newVar` is an empty string. I'm not sure what the implications of that are (it might be fine), but is probably worth checking.

tests/variables_test.js
```diff
@@ -65,3 +65,17 @@ function test_allVariablesOtherCategory() {
...
+  var blockXML = '<xml><block type="variables_get"><title name="VAR">i</title></block></xml>';
+  Blockly.Xml.domToBlockSpace(Blockly.mainBlockSpace, Blockly.Xml.textToDom(blockXML));
+
+  Blockly.Variables.renameVariable("i", "z", Blockly.mainBlockSpace);
+  assertEquals("Variable renamed to 'z'", 'z', Blockly.Variables.allVariables()[0]);
+  Blockly.Variables.renameVariable("z", "", Blockly.mainBlockSpace);
+  assertEquals("Cannot rename var to empty", 'z', Blockly.Variables.allVariables()[0]);
+
+  goog.dom.removeNode(container);
+}
```
Bjvanminnen: yay tests!

joshlory:
Ah good catch, fixing the `setText` issue on line 156.

PR #157, opened 2018-11-16
Add clampedNumberValidator for input fields

islemaster:
Part of the solution for https://github.com/code-dot-org/dance-party/issues/321

Allows us to have a freeform text input in a block that can be constrained to accepting numeric input within a specified range.  The range can have a lower bound, and upper bound, or both.  The corresponding input element will be given the "number" type by default so the student cannot even type non-numeric characters into the field.  Invalid or blank input results in zero, unless zero is outside the given bounds in which case the zero is clamped to the nearest boundary.

### DIFF
--- a/core/ui/fields/field_textinput.js
+++ b/core/ui/fields/field_textinput.js
@@ -148,7 +148,10 @@ Blockly.FieldTextInput.prototype.showEditor_ = function() {
   var div = Blockly.WidgetDiv.DIV;
   // Create the input.
   var htmlInput = goog.dom.createDom('input', 'blocklyHtmlInput');
-  if (this.changeHandler_ === Blockly.FieldTextInput.numberValidator) {
+  if (
+    (this.changeHandler_ === Blockly.FieldTextInput.numberValidator) ||
+    (this.changeHandler_ && this.changeHandler_.validatorType === 'clampedNumberValidator'))
+  {
     htmlInput.setAttribute('type', 'number');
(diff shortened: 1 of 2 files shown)


islemaster:
See it in action on https://github.com/code-dot-org/code-dot-org/pull/26124

core/ui/fields/field_textinput.js
```diff
@@ -148,7 +148,10 @@ Blockly.FieldTextInput.prototype.showEditor_ = function() {
   var div = Blockly.WidgetDiv.DIV;
   // Create the input.
   var htmlInput = goog.dom.createDom('input', 'blocklyHtmlInput');
-  if (this.changeHandler_ === Blockly.FieldTextInput.numberValidator) {
+  if (
+    (this.changeHandler_ === Blockly.FieldTextInput.numberValidator) ||
+    (this.changeHandler_ && this.changeHandler_.validatorType === 'clampedNumberValidator'))
```
Hamms: I don't love that we're introducing inconsistency with the way you identify a `changeHandler`; can we create an issue to track improving this?

islemaster: :+1: That's a good idea.  I messed around with identifying these by function name or prototype instead of by identity, but decided in the end it'd be too clever by half and I should just add a property.  I'd love to revisit this and make it make sense.

https://github.com/code-dot-org/blockly/issues/158

PR #159, opened 2018-11-17
Don't skip `setIsUnused` when main workspace is readonly

joshlory:
Fixes https://github.com/code-dot-org/dance-party/issues/417, generation of code for non-top blocks on share pages.

Originally added in https://github.com/code-dot-org/code-dot-org/commit/df02581e0c3e3b2fa14f068ed942ec9f3013c6f3, @Hamms can you provide more context for why this case was excluded originally?

### DIFF
--- a/core/ui/block.js
+++ b/core/ui/block.js
@@ -1817,7 +1817,6 @@ Blockly.Block.prototype.setIsUnused = function(isUnused) {
         this.isUserVisible() &&
         this.type !== 'functional_definition' &&
         Blockly.mainBlockSpace &&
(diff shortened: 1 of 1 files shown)


joshlory:
@Hamms and I chatted offline, it seems that the main place we embed blocks doesn't use the main block space ([link](https://studio.code.org/s/algebra/stage/2/puzzle/12)).  We're also less worried since the unused overlay is hidden by default, until the user clicks the "Run" button.

There's still a chance we're using readonly block spaces in a weird way, but we think the risk is low for HoC.

Issue #148, opened 2018-09-14
Scrollable dropdowns not discoverable


jakebrbell:
When there is a dropdown with a lot of options on a block, it isn't apparent that you can scroll and see more options. One example of this is from the [Minecraft: Hero's Journey activity](https://studio.code.org/s/hero/stage/1/puzzle/12). 

<img width="259" alt="screen shot 2018-09-14 at 3 21 40 pm" src="https://user-images.githubusercontent.com/12363478/45577721-37de6f00-b832-11e8-9404-9e39d0bfb49c.png">

Some possible solutions:

- Make the scrollbar always visible, so it's clear to the user that it is scrollable. 
- Show a small arrow at the bottom (this would especially be helpful for touch users).
- Make the last viewable item on the list only half shown so it's clear that there are more options.

Separately, it might also be helpful if the dropdown menu took up more space (height) on the screen to reduce times when the user would have to scroll.

janhink:
I second this, even three years later this issue still persists. On Chrome the scrollbar isn't visible so the users don't know they can scroll to see the other options. On Vivaldi the scrollbars are there. So this is a browser specific issue.

PR #161, opened 2018-11-19
Increase the dropdown height to enable a "peek" at the next option again

breville:
[Years ago](https://github.com/code-dot-org/code-dot-org/pull/1542) I set the `max-height` of a blockly dropdown to be `250px` so that we would get a "peek" at the next option, if there were one, so that users would know it was scrollable on systems that don't have scrollbars.

Over time, this size seems to no longer show a peek, and so here's a change to bump it to `265px`:

screenshot 2018-11-20 09 02 47

This change has broad impact, so we should consider its adoption carefully.

This fixes https://github.com/code-dot-org/dance-party/issues/260.

### DIFF
--- a/core/ui/fields/field_dropdown.js
+++ b/core/ui/fields/field_dropdown.js
@@ -135,7 +135,7 @@ Blockly.FieldDropdown.prototype.showEditor_ = function(container) {
     this.sourceBlock_.getSaturation() * 100 + '%, ' +
     this.sourceBlock_.getValue() * 100 + '%' + ', 0.5)';
   menuDom.style.overflowY = "auto";
-  menuDom.style["max-height"] = "250px";
+  menuDom.style["max-height"] = "265px";
 
   this.positionWidgetDiv();
 


Bjvanminnen:
Eyes test might be the easiest?

Hamms:
Should we consider as a longer-term fix making the scrollbar be visible?

breville:
@Hamms If we use native scrolling on modern operating systems I think it's normal to not see a scrollbar.

The 265px value was chosen because it is just a little bigger than the existing 250px, and is big enough to get a peek at the next value that it incentivises a scroll.

Hamms:
I'm pretty sure it's only osx that hides the scrollbar; and normal or not, it's clearly causing confusion

breville:
Fair point.  Good news is that peek seems a reasonable mitigation for now!

If we're doing an eyes test I guess I'll wait until blockly is updated in our main repo.  (I did ponder a PhantomJS test in this repo to make sure that the dropdown max height is not close to a multiple of the item height, but that seems a little... obscure. :)

joshlory:
Related bug? https://github.com/code-dot-org/blockly/issues/148

> Separately, it might also be helpful if the dropdown menu took up more space (height) on the screen to reduce times when the user would have to scroll.

PR #163, opened 2018-11-21
Check block connectivity rules before superiority

islemaster:
Will fix https://github.com/code-dot-org/dance-party/issues/62

There's a subtle case we're guarding against here and probably a deeper fix would be good, but as far as I can tell this does solve the case we care about.

An unconnected block (after paste) could bump a large section of connected blocks, which seems like behavior we never want to use. Now we use the new rules checking whether blocks are connected before bumping first, which gives nicer behavior.

Worse, in a few cases this could lead to an infinite loop of block-bumping:

infinibump

I managed to write a unit test that reproduces this behavior with stock blocks, and after this fix it no longer does:

peek 2018-11-21 14-49

The infinite loop potential was introduced in https://github.com/code-dot-org/blockly/pull/128 so I checked to make sure I'm preserving the desired new behaviors from that PR.

peek 2018-11-21 14-46

And here's the fix working with the known broken case in dance-party:

peek 2018-11-21 14-55

### DIFF
--- a/tests/block_space_test.js
+++ b/tests/block_space_test.js
@@ -72,6 +72,57 @@ function test_bumpNeighbours() {
   goog.dom.removeNode(container);
 }
 
+function test_noInfiniteBumpOnPaste() {
+  var container = Blockly.Test.initializeBlockSpaceEditor();
+
+  // This configuration is known to generate an infinite loop of blocks bumping each other
+  // after the changes in https://github.com/code-dot-org/blockly/pull/128
+  // See https://github.com/code-dot-org/dance-party/issues/62 as an example of this in
+  // the wild.
+  var blockXML = '<xml>' +
+    '<block type="controls_repeat">' +
+      '<statement name="DO">' +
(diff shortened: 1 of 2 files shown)


tests/block_space_test.js
```diff
@@ -72,6 +72,57 @@ function test_bumpNeighbours() {
...
+  var originalWhileBlock = ifBlock.getChildren()[0];
+  assertEquals('controls_whileUntil', originalWhileBlock.type);
+
+  // Store the original while-block location
+  var originalLocation = originalWhileBlock.getRelativeToSurfaceXY();
+
+  // Copy and paste the while-block
+  Blockly.BlockSpaceEditor.copy_(originalWhileBlock);
+  var clipboardXml = Blockly.Xml.textToDom(Blockly.clipboard_);
+  Blockly.mainBlockSpace.paste(clipboardXml);
```
islemaster: Before my change, this line would repro the maximum stack depth exceeded error in unit tests.

PR #166, opened 2018-11-27
don't swallow backspace/delete on phone inputs

davidsbailey:
fixes (5) from https://github.com/code-dot-org/dance-party/issues/462 . applies to any finish dialog on blockly levels, including /s/dance, /s/artist, and many other HoC scripts.


davidsbailey:
screen shot 2018-11-27 at 3 46 05 pm

PR #167, opened 2018-11-28
fix reference error positioning color menu in RTL

davidsbailey:
fixes https://github.com/code-dot-org/dance-party/issues/508

### DIFF
--- a/core/ui/fields/field_colour.js
+++ b/core/ui/fields/field_colour.js
@@ -163,8 +163,9 @@ Blockly.FieldColour.prototype.positionWidgetDiv = function() {
     xy.x += borderBBox.width;
   }
   xy.y += borderBBox.height - 1;
-  if (Blockly.RTL) {
-    xy.x -= div.offsetWidth;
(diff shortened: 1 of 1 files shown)


core/ui/fields/field_colour.js
```diff
@@ -163,8 +163,9 @@ Blockly.FieldColour.prototype.positionWidgetDiv = function() {
     xy.x += borderBBox.width;
   }
   xy.y += borderBBox.height - 1;
-  if (Blockly.RTL) {
-    xy.x -= div.offsetWidth;
+  var div = Blockly.WidgetDiv.DIV;
```
davidsbailey: copied from https://github.com/code-dot-org/blockly/blob/694b7d7624536d1997dd7f0bc1f29c5e47c510dd/core/ui/fields/field_colour.js#L111-L113, which is the function calling this one in the call stack from the bug report.

davidsbailey:
menu now positioned properly:
screen shot 2018-11-28 at 1 38 00 pm

dance puzzle 8 can also be passed now.

PR #170, opened 2019-04-04
Add support for Continuous Integration Builds in Drone

sureshc:
Migrate this repository's Continuous Integration service from CircleCI to self-hosted Drone: https://codedotorg.atlassian.net/browse/INF-95

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -41,3 +41,8 @@ trigger:
   - master
   event:
   - pull_request
+---
+kind: signature
+hmac: 6bce0d8194a8e9cfd42bd173f7cdf5969a54496e8aaa10b5532b6f6d765a2009
+
+...
--- a/.drone.yml
+++ b/.drone.yml
@@ -4,7 +4,7 @@ name: default
 
 steps:
 - name: build-and-test
-  image: node/node
+  image: node:8.6.0
   pull: always
   commands:
    - npm install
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -0,0 +1,43 @@
+---
+kind: pipeline
+name: default
+
+steps:
+- name: restore-cache
```
uponthesun: It seems like caching saves about a minute for the build? I think it'd be fine to not have caching to simplify the config file, thoughts?

sureshc: Agreed.  I haven’t figured out how to cache `node_modules` based on the hash of `package.json`, which is what our CircleCI build currently does for this project.  I think we could just eliminate the caching.

sureshc:
@uponthesun Can you show me how to get our drone server to sign this file?  Do I install Drone command line tools and the configure them to point to our drone server before running the signing tool?

sureshc:
@joshlory I had to configure our codecov token to get the upload of the coverage report working (codecov does not require this for public repositories being built within CircleCI, but in Drone the token is required).  I’m not clear why the report is not being published to the GitHub Pull Request.  I could use a moment of your time to diagnose that.

uponthesun:
@sureshc Yea, that's right. You can find the configuration values for the CLI by clicking on your user > user settings on the top right on the drone site.

PR #173, opened 2019-04-18
Duplicate block on drag from specified parent block

epeach:
semipermanent

Part of our prototype for 'mini-toolboxes': https://docs.google.com/document/d/1VV9qHE0qjvas1o8-7zavDpkrbxG9gUmCFRas3Xlwtx4/edit#heading=h.8u7mwfss3758

This changes allows us to indicate a parent block, that when the current block is attached to, the current block will duplicate itself on drag and the duplicated block will continue the drag event (see gif above). 

Example of how it could be called from cdo repo: 
Screenshot from 2019-04-18 10-09-11

### DIFF
--- a/core/ui/block.js
+++ b/core/ui/block.js
@@ -717,10 +717,9 @@ Blockly.Block.prototype.onMouseDown_ = function(e) {
     return;
   } else {
     // Left-click (or middle click)
-    // If this is connected to the parent from which it should duplicate on drag,
-    // duplicate the block, pass the click event to the duplicated block, and
-    // return from this block's click event
-    if(this.copyOnDrag_ && this.getParent() && (this.getParent().type === this.copyOnDrag_)){
+    // If the block should duplicate on drag, duplicate the block, pass the click event
+    // to the duplicated block, and return from this block's click event
+    if(this.shouldCopyOnDrag()){
       let dup = this.duplicate_();
       dup.setParentForCopyOnDrag(null);
       dup.onMouseDown_(e);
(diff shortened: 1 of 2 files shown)


joshlory:
Cool! Heads up @mikeharv.

How do I tell a clone-able block apart from a normal argument block? If I drop a different block int he socket, does it detach the "toolbox" block?

epeach:
@joshlory Good question! At the moment there isn't any visual distinction. And dropping a different block in the socket does detach the toolbox block (which de-activates the duplicate feature for the displaced toolbox block, until it is replaced in the parent block). This is definitely something we'd want to address if we decide to move beyond the prototype stage with this feature, but I'm not sure it's needed for an initial proof of concept. Happy to change it though, if we determine those changes are needed at this stage.

joshlory:
Ah got it! Remaining UI work is captured in https://codedotorg.atlassian.net/browse/STAR-201.

core/ui/block.js
```diff
@@ -717,6 +717,14 @@ Blockly.Block.prototype.onMouseDown_ = function(e) {
     return;
   } else {
     // Left-click (or middle click)
+    // If the block should duplicate on drag, duplicate the block, pass the click event
+    // to the duplicated block, and return from this block's click event
+    if(this.shouldCopyOnDrag()){
```
joshlory: Wondering if we can de-dup this against the code that duplicates blocks when they're dragged out of a flyout toolbox?

joshlory: Tiny style nit: space after `if`.

joshlory: And before `{` (two more below).

PR #177, opened 2019-06-13
Remove "checkbox" field type

joshlory:
Hasn't been used in any CDO blocks (to my knowledge). Not currently used in any blocks: https://github.com/code-dot-org/code-dot-org/search?q=FieldCheckbox


joshlory:
This will be another difference from https://github.com/google/blockly but I think we're ok diverging further (for our long-term-stable fork).

PR #181, opened 2019-07-09
Rename dropdown options for renaming variables

ajpal:
Do not submit until https://github.com/code-dot-org/code-dot-org/pull/29621 is merged.

The variable dropdowns as-is are consistently a point of confusion for users, so rename the options to make it more clear what each option does.

Old:
dropdown-old

New:
dropdown-new


ajpal:
Closing and will open a new PR with this built as a custom option that we can set from Sprite Lab

PR #183, opened 2019-07-23
Add i18n strings for variable dropdown

ajpal:
Adds strings to all languages for #182


bethanyaconnor:
It was suspicious to me that there was an odd number of files but en_gb is only in there once so I think that's okay (we don't support en_gb and it tends to become very intertwined with any en files anyway). Thanks for doing this and sorry we don't have better tools. :/

PR #191, opened 2019-10-17
Remove dependency on `goog.async.AnimationDelay`, use RAF instead

joshlory:
Start pulling out Clojure library dependencies, to make it easier to port this code to upstream Blockly.


islemaster (commented):
Yeah, that looks good.  Does Blockly have tests for this?

PR #192, opened 2019-10-17
counter_for loop block: Add new i18n string to all locales

nkiruka:
A one-off script was written to add a new i18n string used in the counter_for loop block to all locales.  This ensures that the counter_for loop block is not broken when viewed in a non-US locale.  Follow up work is to eliminate this manual process.


bethanyaconnor:
Some questions from someone who has been out of the loop:
1. Will these strings be added to the correct source files so that they will persist between i18n syncs?
2. I think these strings should be added to the apps i18n files not to the files in the i18n directory?
3. Can you share the code you used to do this? I would love to be able to look back on these PRs and see what to do to solve these issues in the future!

nkiruka:
@bethanyaconnor 
Here are responses to your questions:
1.  Yes, these strings will be added to the correct source files.
2.  When I run this [script](https://github.com/code-dot-org/blockly/blob/master/i18n/codeorg-messages.sh) it will add the strings to apps i18n
3.  Sure, I will update PR to include script.

i18n/add_i18n_strings_to_locales.py
```diff
@@ -0,0 +1,19 @@
...
+for locale in locales:
+    path = "%s/%s/core.json" % (locales_dir, locale)
+    lines = open(path, 'r').readlines()
+
+    with open(path, 'w') as output:
+        done = False
+        for line in lines:
+            segments = line.split('"')
+            if (len(segments) > 1 and segments[1] > "CONTROLS_FOR_HELPURL" and not done):
+                output.write('  "CONTROLS_FOR_INPUT_COUNTER": "for %1 from %2 to %3 count by %4",\n')
```
Hamms: hmm, given that this script is very specific to just this string, it might not actually be appropriate to include it in the repo.

Rather, how about we say that part of your investigation into why this is necessary should include a comparable - but generalized - script.

nkiruka: Good point!  thanks

PR #199, opened 2020-02-05
Fix AngleHelper for Firefox

ajpal:
`e.offsetX/Y` was giving unpredictable values in Firefox, which was causing the flicker effect. According to https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/offsetX, it's an experimental feature anyways, so switching to always use `e.clientX/Y`, which seems more stable.
Also, moved the bounding rect to an instance variable to avoid repeating the expensive computation unnecessarily since the size of the dropdown does not ever change.

Before
Feb-04-2020 16-22-45

After
Feb-04-2020 16-23-26

[jira](https://codedotorg.atlassian.net/browse/STAR-650)

### DIFF
--- a/core/ui/fields/angle_helper.js
+++ b/core/ui/fields/angle_helper.js
@@ -52,6 +52,7 @@ Blockly.AngleHelper = function(direction, opt_options) {
   this.arc_ = null;
   this.circle_ = null;
   this.svg_ = null;
+  this.rect_ = null;
   this.variableLine_ = null;
 
   this.animationInterval_ = null;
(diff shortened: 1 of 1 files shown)


ajpal:
> > Also, moved the bounding rect to an instance variable to avoid repeating the expensive computation unnecessarily since the size of the dropdown does not ever change.
> 
> Can you confirm that the _position_ of the dropdown also does not ever change? It's not clear to me exactly when we create new AngleHelpers; I worry about what will happen if the block is moved between invocations.

If you move the block, the dropdown automatically closes, and then we would recalculate the bounding rect if you re-open the dropdown.

PR #203, opened 2020-03-19
Fix block_svg for firefox

ajpal:
`getBBox` does not work with hidden elements in Firefox. https://bugzilla.mozilla.org/show_bug.cgi?id=528969

This is a targeted fix to fix https://studio.code.org/s/coursec-2019/stage/16/puzzle/6, but we should probably look into whether this is causing more widespread issues with Blockly labs on Firefox


madelynkasula:
do we need to do something manual to bump blockly to 3.5.8? i see we're pulling that version into code-dot-org, but i'm not sure how that version gets published

ajpal:
I ran `npm version patch` from the `master` branch of blockly locally, which created https://github.com/code-dot-org/blockly/commit/4250a4249c64037b8f1f4e95346f17cfd09e0a96

PR #204, opened 2020-03-23
Angle Helper UI refresh

ajpal:
Before:
image

After:
Chrome
image
Safari
image
Firefox
image


jmkulwik:
Works great in Edge and IE as well

core/ui/fields/angle_helper.js
```diff
@@ -153,7 +144,8 @@ Blockly.AngleHelper.prototype.init = function(svgContainer) {
     var isOnPrimaryHalf = this.turnRight_ ? angle < 180 : angle > 180;
     Blockly.createSvgElement('line', {
       'stroke-linecap': 'round',
-      'stroke-opacity': isOnPrimaryHalf ? 1 : 0.3,
```
Hamms: can you confirm that

a) we no longer want this "primary half" functionality and
b) we can remove the `isOnPrimaryHalf` variable

ajpal: Yes the two tones of tick marks was intentionally removed. Removed the unused variable, thanks for catching

PR #206, opened 2020-03-25
[DO-NOT-MERGE] Add "block id" as Attribute in XML Serialization

mhw32:
Example:

```
<xml><block type="when_run" deletable="false" movable="false" blockid="6"><next><block type="maze_turn" blockid="7"><title name="DIR">turnRight</title><next><block type="maze_nectar" blockid="8"><next><block type="controls_repeat" blockid="9"><title name="TIMES">5</title><statement name="DO"><block type="maze_turn" blockid="10"><title name="DIR">turnLeft</title><next><block type="maze_nectar" blockid="11"></block></next></block></statement></block></next></block></next></block></next></block></xml>
```

### DIFF
--- a/core/utils/xml.js
+++ b/core/utils/xml.js
@@ -155,6 +155,9 @@ Blockly.Xml.blockToDom = function(block, ignoreChildBlocks) {
     element.setAttribute('id', block.htmlId);
   }
 
+  // Add ID as an attribute
+  element.setAttribute('blockid', block.id);
+
   // Don't follow connections if we're ignoring child blocks
   if (block.nextConnection && !ignoreChildBlocks) {
     var nextBlock = block.nextConnection.targetBlock();


core/utils/xml.js
```diff
@@ -155,6 +155,9 @@ Blockly.Xml.blockToDom = function(block, ignoreChildBlocks) {
     element.setAttribute('id', block.htmlId);
   }
 
+  // Add ID as an attribute
+  element.setAttribute('blockid', block.id);
```
Hamms: We don't want to do this for everything; in particular, we rely pretty heavily on blockly source deduplication to optimize our storage of student solutions. Could you add an optional argument to situationally enable this?

mhw32: Of course! in fact i will add a DO NOT MERGE in this title. I'm opening two pull requests mainly as a way for you to keep track of changes i make. I expect we may even deprecate both requests for the production version fo code

mhw32: Revisiting this, i think it is only used for highlighting but we don't need it for now. Will close this for now until further notice.

PR #207, opened 2020-03-25
Angle Picker UI Refresh

ajpal:
Enable rotating the baseline as well as picking an angle
Mar-25-2020 13-52-06

UI for this is probably not 100% finalized, but wanted to at least start the review process for the functional changes.

Tested locally in Chrome, Firefox, and Safari on Mac

### DIFF
--- a/core/ui/fields/angle_helper.js
+++ b/core/ui/fields/angle_helper.js
@@ -31,24 +31,28 @@ Blockly.AngleHelper = function(direction, opt_options) {
   
   this.turnRight_ = direction === 'turnRight';
   this.lineColour_ = '#4d575f';
-  this.handleR_ = 7.5;
+  this.pickerHandleR_ = 7.5;
   this.draggingHandle_ = false;
   this.draggingBaseline_ = false;
   this.strokeWidth_ = 3;
   this.center_ = new goog.math.Vec2(this.width_ / 2, this.height_ / 2);
+  this.radius_ = new goog.math.Vec2(Math.min(this.height_, this.width_) / 2 - this.pickerHandleR_ - this.strokeWidth_, 0);
+  
   this.baselineAngle_ = 0;
+  this.baselineHandle_ = null;
+  this.baselineHandleR_ = 4;
+  this.baselineHandleCenter_ = this.center_.clone().add(new goog.math.Vec2(this.radius_.x + this.baselineHandleR_, 0));
 
-  this.radius_ = new goog.math.Vec2(Math.min(this.height_, this.width_) / 2 - this.handleR_ - this.strokeWidth_, 0);
-
-  this.handleCenter_ = this.center_.clone().add(new goog.math.Vec2(this.radius_.x - this.handleR_, 0));
-  this.handleCenter_ = goog.math.Vec2.rotateAroundPoint(
-    this.handleCenter_,
+  this.pickerHandleCenter_ = this.center_.clone().add(new goog.math.Vec2(this.radius_.x - this.pickerHandleR_, 0));
+  this.pickerHandleCenter_ = goog.math.Vec2.rotateAroundPoint(
+    this.pickerHandleCenter_,
     this.center_,
     goog.math.toRadians(this.turnRight_ ? this.angle_ : -this.angle_)
   );
 
   this.arc_ = null;
-  this.handle_ = null;
+  this.pickerHandle_ = null;
+  this.baselineHandle_ = null;
   this.svg_ = null;
   this.rect_ = null;
   this.pickerLine_ = null;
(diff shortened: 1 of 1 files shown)


jmkulwik:
If this isn't a finished UI, would you mind making the PR into a feature branch? That way we can still publish from master if we need an urgent bug fix.

ajpal:
I won't merge this PR until the UI is agreed upon, but I think the plan is to get this done within the next day or two, so I think leaving it as a PR against master makes sense.

core/ui/fields/angle_helper.js
```diff
@@ -183,52 +181,74 @@ Blockly.AngleHelper.prototype.init = function(svgContainer) {
  * onUpdate callback
  */
 Blockly.AngleHelper.prototype.update_ = function() {
-  this.circleCenter_ = goog.math.Vec2.rotateAroundPoint(
-    this.center_.clone().add(this.lineLength_),
+  this.baseLine_.setAttribute('transform', 'rotate(' + this.baselineAngle_ + ', ' + this.center_.x + ', ' + this.center_.y + ')');
```
jmkulwik: Use string interpolation here. i.e. 
```
`rotate(${this.baselineAngle_}, ${this.center_.x}, ${this.center_.y})`
```

jmkulwik: (and below)

ajpal: Isn't that an ES6 feature?

jmkulwik: 😢 good point

jmkulwik:
Looks great. My only comment is to add in some comments. Especially in areas where you are setting the angles.

Hamms:
have we considered adding a "handle" you can grab to manipulate the rotation? As-is, it seems super unclear to me that rotation is an option, and entirely too easy to accidentally rotate

core/ui/fields/angle_helper.js
```diff
@@ -183,52 +181,74 @@ Blockly.AngleHelper.prototype.init = function(svgContainer) {
...
+  var x = e.clientX - this.rect_.left;
+  var y = e.clientY - this.rect_.top;
+  var mouseLocation = new goog.math.Vec2(x, y);
+  if (goog.math.Vec2.distance(this.handleCenter_, mouseLocation) < this.handleR_) {
+    this.draggingHandle_ = true;
     return;
   }
+  var angle = goog.math.angle(this.center_.x, this.center_.y, x, y);
+  var distanceFromCenter = goog.math.Vec2.distance(this.center_, mouseLocation);
+  if (Math.abs(this.baselineAngle_ - angle) < 10 && distanceFromCenter < (this.radius_.magnitude() + 10)) {
```
Hamms: what's the significance of `10` here?

ajpal:
Yes, the UI for the rotation should be coming later today, I just wanted to get a head start on the tricky pieces of the implementation before everything was 100% finalized

ajpal:
New UI:

Mar-26-2020 10-06-25

core/ui/fields/angle_helper.js
```diff
@@ -40,7 +40,7 @@ Blockly.AngleHelper = function(direction, opt_options) {
 
   this.radius_ = new goog.math.Vec2(Math.min(this.height_, this.width_) / 2 - this.handleR_ - this.strokeWidth_, 0);
 
-  this.handleCenter_ = this.center_.clone().add(this.radius_)
+  this.handleCenter_ = this.center_.clone().add(new goog.math.Vec2(this.radius_.x - this.handleR_, 0));
```
jmkulwik: Tried out the new handle on a touch screen (iPad). I could only get it to move ~1/5 times due to the smaller size.

core/ui/fields/angle_helper.js
```diff
@@ -183,52 +200,83 @@ Blockly.AngleHelper.prototype.init = function(svgContainer) {
...
-  this.dragging_ = true;
-};
-
-Blockly.AngleHelper.prototype.updateDrag_ = function(e) {
-  if (!this.dragging_) {
+Blockly.AngleHelper.prototype.startDrag_ = function(e) {
+  var x = e.clientX - this.rect_.left;
+  var y = e.clientY - this.rect_.top;
+  var mouseLocation = new goog.math.Vec2(x, y);
+  if (goog.math.Vec2.distance(this.pickerHandleCenter_, mouseLocation) < this.pickerHandleR_) {
```
jmkulwik: Requiring the mouse to be within the bounds of the picker handle may make usage on a touch screen more difficult.

Hamms: oooooh yeah if I recall correctly that's why we added the "touch background to set angle" functionality in the first place.

With the new handle for moving the background, is it still necessary to remove that?

jmkulwik:
Took a video of usage on an iPad: https://photos.app.goo.gl/xKVxrHrJdHh596zXA
Couple problems:
1. the angle picker no longer jumps to where you drag on the screen (you need to start your drag from within the circle)
2. the baseline picker is too small to reasonably be able to drag on touch

ajpal:
Closing because product doesn't want this change anymore. Can reopen if needed when design is finalized.

PR #215, opened 2020-04-22
Angle Helper UI refresh

ajpal:
Mostly the same as https://github.com/code-dot-org/blockly/pull/207 but with UI that is now finalized by Product.
Clicking anywhere except inside the small gray knob will start dragging the picker line. Clicking inside the small gray knob will start rotating the background.
The small gray knob is likely a pretty small target for touch devices, however Marina said this was the right tradeoff- it should be very easy to move the picker line, and it's okay if it's harder to rotate the background.
Apr-21-2020 17-01-28


core/ui/fields/angle_helper.js
```diff
@@ -195,10 +214,23 @@ Blockly.AngleHelper.prototype.init = function(svgContainer) {
  * onUpdate callback
  */
 Blockly.AngleHelper.prototype.update_ = function() {
+  if (this.enableBackgroundRotation_) {
+    this.background_.line.setAttribute('transform',
+    'rotate(' + this.background_.angle + ', ' + this.center_.x + ', ' + this.center_.y + ')');
+    for (var i = 0; i < this.background_.ticks.length; i++) {
+      var angle = (15 * i + this.background_.angle) % 360;
```
madelynkasula, line 221: should `15` be a constant somewhere? it looks like we use it in a few different places in this file

core/ui/fields/angle_helper.js
```diff
@@ -195,10 +222,23 @@ Blockly.AngleHelper.prototype.init = function(svgContainer) {
  * onUpdate callback
  */
 Blockly.AngleHelper.prototype.update_ = function() {
+  if (this.enableBackgroundRotation_) {
```
madelynkasula, line 225: there's a lot of new logic in this method and in the `startDrag_()` method -- can we cover this with unit tests?

PR #217, opened 2020-04-28
Add a background behind flyout button

ajpal:
Currently for toolbox categories with a flyout button, there's an issue where the button overlaps the blocks as you scroll.
To resolve, we want to just add an opaque background behind the flyout button so that the blocks scroll underneath.
Before:
Apr-28-2020 11-54-43

After:
Apr-28-2020 11-53-23

### DIFF
--- a/core/ui/block_space/flyout.js
+++ b/core/ui/block_space/flyout.js
@@ -574,6 +574,12 @@ Blockly.Flyout.prototype.show = function(xmlList) {
   }
   this.width_ = 0;
   this.reflow();
+  // After filling the flyout, if there's a flyout button, we should 
+  // update the width of the background rectange to fill the width of the flyout
+  var flyoutButton = document.getElementsByClassName('createFunction')[0];
+  if (flyoutButton) {
+    flyoutButton.children[0].setAttribute('width', Math.max(0, this.getMetrics_().viewWidth - 15));
+  }
 
   this.filterForCapacity_();
   this.updateBlockLimitTotals_();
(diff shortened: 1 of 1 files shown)


core/ui/block_space/flyout.js
```diff
@@ -574,6 +574,12 @@ Blockly.Flyout.prototype.show = function(xmlList) {
   }
   this.width_ = 0;
   this.reflow();
+  // After filling the flyout, if there's a flyout button, we should 
+  // update the width of the background rectange to fill the width of the flyout
+  var flyoutButton = document.getElementsByClassName('createFunction')[0];
```
Hamms, line 579: I'd strongly suggest that we have the addButtonToFlyout method save a reference to the created element somewhere, and have the `reflow` method take care of all size adjustments; this approach has significant potential for accidental side effects

boatmarina:
great idea!

core/ui/block_space/flyout.js
```diff
@@ -650,6 +670,9 @@ Blockly.Flyout.prototype.reflow = function() {
         block.flyoutRect_.setAttribute('y', blockXY.y);
       }
     }
+    if (this.flyoutButton_) {
+      this.flyoutButton_.children[0].setAttribute('width', Math.max(0, flyoutWidth - 15));
```
Hamms, line 674: this is a bit nitpicky, but I'd prefer to save a reference directly to the element we're going to be modifying, for two reasons. First, it's not clear here what we're actually modifying (the background, right?), and second calling this `flyoutButton` seems to be inaccurate, given that it's actually the container for the button and what we're modifying is actually a different part of the container than the button itself.

PR #218, opened 2020-05-26
Sprite Lab Pointer Blocks Shadowing Functionality

ajpal:
Pairs with https://github.com/code-dot-org/code-dot-org/pull/34975
This shadowing behavior is *not* general purpose, and will only work for the combination of blocks needed for mini toolboxes for events. That is, there is a clicked sprite pointer block that can shadow the sprite dropdown block in a clicked event block, and there are subject/object pointer blocks that can shadow the sprite dropdowns in a touch event block. If we ever need any other shadowing, that will need to be built separately. 

May-26-2020 15-02-41
May-26-2020 15-02-12

### DIFF
--- a/core/ui/block.js
+++ b/core/ui/block.js
@@ -1505,7 +1505,7 @@ Blockly.Block.prototype.setParent = function(newParent) {
     // Add this block to the new parent's child list.
     newParent.childBlocks_.push(this);
     // If the new block has relational blocks to update, re-render the blocks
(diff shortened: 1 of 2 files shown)


jmkulwik:
Offline conversation: There are currently a couple of scenarios that might seem a little odd from a UI perspective, but will be addressed with some follow-up work

image

Future work: Clear the shadow block when it is removed from its root stack.

PR #219, opened 2020-05-28
Sprite Lab Pointer Blocks Shadowing Functionality

ajpal:
Follow up to https://github.com/code-dot-org/blockly/pull/218 because that one didn't correctly handle when the sprite blocks were moved. 
There are 5 cases where we might need to update previews on shadow blocks:

1. A sprite block is added to an event socket
May-27-2020 17-02-55

2. A sprite block is removed from an event socket
May-27-2020 17-02-31

3. A stack of blocks is added to an event stack
May-27-2020 17-15-18

4. A stack of blocks is removed from an event stack
May-27-2020 17-15-43

5. Value in the sprite dropdown changes
May-27-2020 17-16-58

The last case is sort of a special case because it doesn't change the block tree at all. In the first four cases, though, the process is basically that we need to find all of the shadow blocks in the stack, and update them with the new thumbnail (or remove the thumbnail). This is done by a straightforward tree traversal (`getShadowBlocksInStack`).

For the 5th case, we maintain a list of shadow blocks for each sprite block. So then when the value changes, all we need to do is go through that list and update the previews:
https://github.com/code-dot-org/blockly/blob/b11192fbc9054e0a6b11f20333d9ea78b08d312a/core/ui/fields/field_rectangular_dropdown.js#L239-L246

### DIFF
--- a/core/ui/block.js
+++ b/core/ui/block.js
@@ -1465,10 +1465,30 @@ Blockly.Block.prototype.getChildren = function() {
 };
 
 /**
+ * Traverses the block stack rooted at topBlock and returns a list of all
+ * the blocks in the stack that have blockToShadow_ set.
+ * @param {Blockly.Block} topBlock - root of the block stack to traverse
+ * @return {!Array.<!Blockly.Block>} Array of shadow blocks
+ */
+function getShadowBlocksInStack(topBlock, sourceBlock) {
+  var shadowBlocks = [];
+  var queue = [topBlock];
+  while (queue.length) {
+    var block = queue.shift()
+    if (block.blockToShadow_) {
+      shadowBlocks.push(block)
+    }
+    queue = queue.concat(block.childBlocks_);
+  }
+  return shadowBlocks;
+}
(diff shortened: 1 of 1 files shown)


core/ui/block.js
```diff
@@ -1464,11 +1464,31 @@ Blockly.Block.prototype.getChildren = function() {
   return this.childBlocks_;
 };
 
+/**
+ * Traverses the block stack rooted at topBlock and returns a list of all
+ * the blocks in the stack that have blockToShadow_ set.
+ * @param {Blockly.Block} topBlock - root of the block stack to traverse
+ * @return {!Array.<!Blockly.Block>} Array of shadow blocks
+ */
+function getShadowBlocksInStack(topBlock, sourceBlock) {
```
madelynkasula, line 1473: it doesn't look like the `sourceBlock` param is used -- can we remove it?

ajpal: whooops yeah that's leftover from another idea

PR #220, opened 2020-05-28
Allowed the toolbox to display with a truncated width

jmkulwik:
Prior to this change, the toolbox would expand to fully display all blocks. On small screens, long blocks would cause the toolbox to take up a disproportionate amount of space on the screen. In extreme cases, the toolbox would take up the entire screen and leave no space for the workspace. If we have toolbox categories, this isn't an issue because the toolbox is an overlay that can be dismissed. Without categories, however, this can cause certain labs to become unusable. We address this here by creating a clipping rectangle to define the maximum toolbox space.

- [Task](https://codedotorg.atlassian.net/browse/STAR-1104)
- [Bug](https://codedotorg.atlassian.net/browse/STAR-673)

Example where translation caused the toolbox to take up the entire screen:
https://studio.code.org/s/course2/stage/17/puzzle/11 in Norsk (Bokmål)

Before:
image

After:
image

Example where small screen sizes caused the toolbox to take up most of the screen
https://studio.code.org/s/dance-2019/stage/1/puzzle/4

Before:
Screenshot_20200528-161602_Chrome

After:
Screenshot_20200528-161516_Chrome

Live Resizing:
ToolboxResize2

### DIFF
--- a/core/ui/block_space/block_space_editor.js
+++ b/core/ui/block_space/block_space_editor.js
@@ -357,7 +357,7 @@ Blockly.BlockSpaceEditor.prototype.addFlyout_ = function() {
    */
   this.flyout_ = new Blockly.Flyout(this, true);
   var flyout = this.flyout_;
-  var flyoutSvg = flyout.createDom(false);
+  var flyoutSvg = flyout.createDom(false /*insideToolbox*/, true /*clipWidth*/);
   flyout.init(this.blockSpace, true);
   flyout.autoClose = false;
   // Insert the flyout behind the blockSpace so that blocks appear on top.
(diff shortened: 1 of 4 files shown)


core/ui/block_space/block_space_editor.js
```diff
@@ -357,7 +357,7 @@ Blockly.BlockSpaceEditor.prototype.addFlyout_ = function() {
    */
   this.flyout_ = new Blockly.Flyout(this, true);
   var flyout = this.flyout_;
-  var flyoutSvg = flyout.createDom(false);
+  var flyoutSvg = flyout.createDom(false /*insideToolbox*/, true /*shouldClipWidth*/);
```
Hamms, line 360: (feel free to ignore if this PR isn't actually ready for review)

I'd suggest that if we feel a method signature is so unclear as to require inline comments to be easily understandable, we should update the method signature.

jmkulwik: We're moving toward a convention where we include these comments in cases where there is a literal passed to a method. I just added it by default. 

That being said, you make a great point - this method signature isn't particularly clear. I'll look into updating it.

PR #221, opened 2020-06-03
Only rerender blockspace when shadow blocks change

ajpal:
This fixes the issue that caused this [revert](https://github.com/code-dot-org/code-dot-org/pull/35103), see also this [slack thread](https://codedotorg.slack.com/archives/C1B3PNDL7/p1591129863387900).
Basically the issue was that `newParent.getShadowBlocks()` is always true (`[]` is truthy), so we were rerendering the blockspace (which triggers a resize event) every time `setParent` was called. Updated to only call `this.blockSpace.render()` when blocks are added or removed from a parent with a mini flyout.

Before (Over 13000 resize events!):
image
After (204 resizes) (this is still a lot, but that's known/unrelated tech debt):
image

### DIFF
--- a/core/ui/block.js
+++ b/core/ui/block.js
@@ -1524,10 +1524,6 @@ Blockly.Block.prototype.setParent = function(newParent) {
   if (newParent) {
     // Add this block to the new parent's child list.
     newParent.childBlocks_.push(this);
-    // If the new block has relational blocks to update, re-render the blocks
-    if (newParent.getShadowBlocks()) {
-      this.blockSpace.render();
-    }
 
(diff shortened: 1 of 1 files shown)


ajpal:
> Is there a reason to change the logic here, rather than just fixing the bug by doing something like `if (newParent.getShadowBlocks().length > 0)`?

Yeah, as it turned out, the new parent having shadow blocks wasn't actually the condition that's important here (so the main functionality was relying on the bug of always rerendering regardless of whether there were shadow blocks). What's actually important is finding when stacks are attached or unattached from blocks that have mini flyouts, which the new logic does.

PR #228, opened 2020-06-24
Don't update shadow values on mini toolbox blocks when mini toolbox is closed

ajpal:
There was a rendering issue in some browsers (Firefox, IE, Edge):
image

The root cause is that `getComputedTextLength()` returns 0 if the text element is not visible. So basically, when we updated the object block, the "object" label thought its width was 0, so the preview went on top of it. The solution is to not update the minitoolbox blocks when the flyout is closed, and [this corresponding PR](https://github.com/code-dot-org/code-dot-org/pull/35455) updates them when the flyout opens

This PR also renames `tray` to `isMiniFlyoutOpen`

### DIFF
--- a/core/ui/block_svg/block_svg.js
+++ b/core/ui/block_svg/block_svg.js
@@ -1293,16 +1293,18 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(renderInfo, connectionsXY
     }
     renderInfo.curY += row.height;
   }
-  if (this.block_.tray) {
-    this.block_.miniFlyout.customMetrics = () => ({
-      absoluteTop: renderInfo.curY - 4,
-      absoluteLeft: Blockly.RTL ? -renderInfo.curX : 5,
-      viewWidth: renderInfo.curX - 10
-    });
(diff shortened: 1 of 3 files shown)


core/ui/block_svg/block_svg.js
```diff
@@ -1293,16 +1293,18 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(renderInfo, connectionsXY
     }
     renderInfo.curY += row.height;
   }
-  if (this.block_.tray) {
```
ajpal, line 1296: there's no change to the logic here, I just think this is a little more readable. 
`this.block_.miniFlyout` will be truthy for any block that has a minitoolbox, regardless of whether it's open. `this.block_.tray` (renamed to `this.block_.isMiniFlyoutOpen`) is true if the flyout is open.
so before it was
```
if (flyout is open) {
  show()
} else if (there is a flyout) {
  hide()
}
```
And now it is
```
if (there is a flyout) {
  if (it's open) {
    show()
  } else {
    hide()
  }
}
```

PR #230, opened 2020-06-25
Add ESLint and fix ES5 violations

ajpal:
Right now the lint rule *only* looks for ES5 violations, I'm going to tackle adding more rules in future PRs. Also it doesn't run before you commit yet.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,13 +1,6 @@
 {
-  "parser": "babel-eslint",
   "parserOptions": {
-    "sourceType": "module",
-    "ecmaFeatures": {
-      "jsx": true,
-      "modules": true,
-      "ecmaVersion": 6,
-      "experimentalObjectRestSpread": true
-    }
+    "ecmaVersion": 5
   },
   "plugins": [
     "es5"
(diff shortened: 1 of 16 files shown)


.eslintrc.json
```diff
@@ -0,0 +1,27 @@
+{
+  "parser": "babel-eslint",
```
Hamms, line 2: Will the default parser work for us? If so, I'd suggest using that instead of babel, since the whole point of this is that we aren't actually transpiling anything

.eslintrc.json
```diff
@@ -0,0 +1,27 @@
...
+  "parserOptions": {
+    "sourceType": "module",
+    "ecmaFeatures": {
+      "jsx": true,
+      "modules": true,
+      "ecmaVersion": 6,
+      "experimentalObjectRestSpread": true
+    }
```
Hamms, line 10: This seems off, for a handful of reasons. I'd suggest something like 

```suggestion
  "parserOptions": {  
    "ecmaVersion": 5
```

.eslintrc.json
ajpal: With the default parser we get errors like 
image
instead of 
image
but I guess that's probably fine?

Hamms: oh, interesting. Yeah, it's not ideal but I think that's fine, and definitely better than introducing a babel-sized dependency

PR #232, opened 2020-06-25
IE fixes for minitoolbox

ajpal:
two targeted fixes for IE compatibility. @jmkulwik tested and confirmed that mini toolbox now works in IE.

### DIFF
--- a/core/ui/block.js
+++ b/core/ui/block.js
@@ -264,7 +264,7 @@ Blockly.Block.prototype.initSvg = function() {
     this.miniFlyout = new Blockly.HorizontalFlyout(this.blockSpace.blockSpaceEditor);
     this.miniFlyout.targetBlockSpace_ = this.blockSpace;
     var dom = this.miniFlyout.createDom(true);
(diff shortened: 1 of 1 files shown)


core/ui/block.js
```diff
@@ -275,7 +275,16 @@ Blockly.Block.prototype.initSvg = function() {
  */
 Blockly.Block.prototype.initMiniFlyout = function(blockString) {
   var root = Blockly.Xml.textToDom(blockString);
-  this.miniFlyoutBlocks = root.children;
+  // Use childNodes, not children, for IE compatibility
+  var childNodes = root.childNodes;
+  var blockList = [];
+  for (var i = 0; i < childNodes.length; i++) {
```
jmkulwik, line 281: NIT: may be nice to use `node` rather than `i`

ajpal: hmm, since it's an index (not the node itself) I actually kind of prefer `i`. I think `i` is a really common counter variable for standard for loops.

PR #233, opened 2020-06-26
Render minitoolbox properly in embedded/readonly blockspaces

ajpal:
Small fix to hidden flyout render properly in readOnly blockspaces (ie. _not_ render)
Also disallows dragging blocks out of the mini toolbox if it's readOnly

Before:
image

After:
image

### DIFF
--- a/core/ui/block_space/block_space_editor.js
+++ b/core/ui/block_space/block_space_editor.js
@@ -258,6 +258,14 @@ Blockly.BlockSpaceEditor.prototype.createDom_ = function(container) {
   container.setAttribute('dir', 'LTR');
 
   this.populateSVGEffects_(container);
+  
+  var classes = ['blocklySvg'];
+  if (this.inline_) {
+    classes.push('inline');
+  }
+  if (this.readOnly_) {
+    classes.push('readOnlyBlockSpace');
+  }
 
   // Build the SVG DOM.
   /*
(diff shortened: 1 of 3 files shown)


core/ui/block_space/block_space_editor.js
```diff
@@ -306,6 +306,8 @@ Blockly.BlockSpaceEditor.prototype.createDom_ = function(container) {
     // Add a handler that allows the workspace to bump blocks back into their
     // working area.
     this.addChangeListener(this.bumpBlocksIfNotDragging);
+  } else {
+    Blockly.addClass_(svg, 'readOnlyBlockSpace');
```
Hamms, line 310: nit; I'd suggest writing this as

```javascript
  if (this.isReadOnly()) {
    Blockly.addClass_(svg, 'readOnlyBlockSpace');
  } else {
    // Determine if there needs to be a category tree, or a simple list of
    // blocks.  This cannot be changed later, since the UI is very different.
    this.addToolboxOrFlyout_();
    // Add a handler that allows the workspace to bump blocks back into their
    // working area.
    this.addChangeListener(this.bumpBlocksIfNotDragging);
  }
```

for readability

Hamms: or possibly rewriting the creation above; something like

```javascript
  var svgClasses = ["blocklySvg"];
  if (this.inline_) {
    svgClasses.push("inline");
  }
  if (this.isReadOnly()) {
    svgClasses.push("readOnlyBlockSpace");
  }

  var svg = Blockly.createSvgElement('svg', {
    'xmlns': 'http://www.w3.org/2000/svg',
    'xmlns:html': 'http://www.w3.org/1999/xhtml',
    'xmlns:xlink': 'http://www.w3.org/1999/xlink',
    'version': '1.1',
    'class': svgClasses.join(" ")
  }, null);
  this.svg_ = svg;
```

PR #234, opened 2020-06-29
Add prettier and apply changes

ajpal:
There are 535 errors that couldn't be fixed automatically, but I'll do that in the next PR just to keep the manually-edited code diff separate from this massive generated one.


ajpal:
> what's the best way for us to tell if something's broken after this change? would it be when we integrate the new blockly version into code-dot-org?

Yeah probably the CDO PR to bump blockly would fail drone. I'm less worried about this PR than the next one, though. I think prettier's pretty good about guaranteeing that changes are no-op.

PR #253, opened 2020-10-16
[Mini Toolbox] Set miniflyoutopen in block xml

ajpal:
Levelbuilders want to be able to set the miniflyout to default to open in instructions. Adding it to the block XML will allow this. Incidentally, it means the miniflyout state will also persist when you re-open a project, which is an added benefit.


madelynkasula (approved):
LGTM!

ajpal:
Closing because this change is no longer needed

PR #254, opened 2020-11-20
Add support for IDs in block titles

ajpal:
Add an ID field to titles that persists across a roundtrip through XML
Before:
image

After:
image

This will be used for Sprite Lab shared behaviors so that we can generate functions by the id of the block, not the text on the block. This will enable us to translate the block text without changing the underlying function.


madelynkasula (commented):
are there any unit tests we can add here? in particular seems like the changes in generators/javascript/procedures.js and core/utils/xml.js should get coverage

madelynkasula (approved):

madelynkasula (approved):

PR #255, opened 2020-11-20
Temporarily disable CodeCov report and sign new Drone script

sureshc:
When the Drone server was rebuilt 2 months ago we lost the CodeCov.io API key being used by Blockly continuous integration builds to publish code coverage reports. Sign the Drone build script so that it is authenticated by the new Drone server and temporarily disable CodeCov.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -18,9 +18,10 @@ steps:
 
 trigger:
   branch:
-  - main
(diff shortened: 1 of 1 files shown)


ajpal (approved):
lgtm, thanks for taking care of this

PR #259, opened 2020-12-15
Adding the Crowdin Pseudo-language for easier translations

daynew:
We are adding a new pseudo language generated by Crowdin. By applying these translations, translators can use the Crowdin plugin to translate blocks while looking at the page they are on.

I followed the example of adding the Mongolian language: https://github.com/code-dot-org/blockly/pull/201

[jira](https://codedotorg.atlassian.net/browse/FND-1381)


annaxuphoto (approved):
LGTM

PR #260, opened 2021-01-11
Fix behavior renaming

ajpal:
[Bug report](https://www.loom.com/share/f289f0c9292247ee941ca8ded4666282)
There were two small issues:
1. If you rename a shared behavior, we don't change the block id, just the text on the block. However, in the modal editor, we were showing the block id, so it didn't match the expected name. 
Before
image

After
image

2. When you rename a user-created behavior, we *do* change the id on the block definition block. However, we need to make sure we also change the associated id on all the existing caller blocks. This resulted in errors the next time you try to open the behavior definition. The fix in the blockly repo is to pass `userCreated` through to the changeHandler when the definition block is renamed. There will be a corresponding change [here](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/p5lab/spritelab/blocks.js#L541) to make it so that we set the id on the caller block if the behavior is user-created
Before
image
After
image

### DIFF
--- a/core/ui/function_editor.js
+++ b/core/ui/function_editor.js
@@ -197,6 +197,7 @@ Blockly.FunctionEditor.prototype.openAndEditFunction = function(functionName) {
   if (!targetFunctionDefinitionBlock) {
     throw new Error("Can't find definition block to edit");
   }
+  var functionDisplayName = targetFunctionDefinitionBlock.getTitleValue('NAME');
 
   this.show();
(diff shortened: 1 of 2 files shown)


madelynkasula (approved):
thank you for the quick fix! are there any tests we can add here to prevent a regression? (i don't think that should block this from going out -- it can be follow-up work)

PR #262, opened 2021-03-23
FND-22: added a condition for positioning inline blocks in RLT context

marcd123:
FND-22: https://codedotorg.atlassian.net/browse/FND-22?atlOrigin=eyJpIjoiYjExMmNlZGQ0ZjI2NGExZThiNTNiYzQwZGYwZTRjY2QiLCJwIjoiaiJ9

Most blocks were being positioned correctly in RTL context, however, inline blocks found in the instructions pane were often slightly offset and clipping into the regular text. Some blocks are so offset they aren't even visible.

The viewWidth is used to calculate offset in a regular context, but for inline we must use the width of the block itself.

Before:

image

After:

image

### DIFF
--- a/core/utils/xml.js
+++ b/core/utils/xml.js
@@ -263,6 +263,10 @@ Blockly.Xml.domToBlockSpace = function(blockSpace, xml) {
   //  otherwise, the block "flows" with the other blocks from top to
   //  bottom. Any block positioned absolutely with Y does not influence
   //  the flow of the other blocks.
+  var cursor = {
+    x: paddingLeft,
+    y: paddingTop
+  };
 
   var positionBlock = function(block) {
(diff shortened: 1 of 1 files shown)


core/utils/xml.js
```diff
@@ -277,6 +274,15 @@ Blockly.Xml.domToBlockSpace = function(blockSpace, xml) {
 
     var heightWidth = block.blockly_block.getHeightWidth();
 
+    var cursor = {
+      x: !Blockly.RTL
```
ajpal, line 278: nit: I think this nested ternary is pretty hard to read. Could be clearer using `if`s instead?

marcd123: I simplified the ternary in my most recent commit

ajpal (approved):

PR #263, opened 2021-03-24
Don't crop images in field image dropdowns

ajpal:
Unfortunately, I can't find any git history (just [Anybody can learn](https://github.com/code-dot-org/blockly/commit/26a17fd70dc273851d359bed0e24647e94e39eaa)) showing why we had originally wanted `xMidyMid slice`, but I don't think we want this anymore. In sprite lab it leads to suboptimal cropping of images in the sprite dropdown:
Before:
image

After:
image

I checked other places we use the FieldImageDropdown, and it doesn't look like it regresses anything for other labs. I think this is because the rest of our images are already square, so there's no cropping needed.
image
image
image


madelynkasula (approved):
great find!

PR #265, opened 2021-03-24
FND-1377 Flyout subject blocks outside display area in RTL

marcd123:
FND-1377: https://codedotorg.atlassian.net/browse/FND-1377?atlOrigin=eyJpIjoiOTc2NzRjMjk1NDllNDIzZDkzZTU4NzA4MGY5ZmVhZTIiLCJwIjoiaiJ9

In RTL context, the content of the flyout was positioned out to the right of the container it should be in.

This is likely because that is where it sits while hidden. This is not an issue in LTR context because it doesn't need to know the width of the content before positioning each element. In RTL we do need to know the width of the content before it's actually in the DOM.

To fix the issue, if RTL - we calculate the width of the flyout content before setting the X and Y offset positions.

Before:

image

After:

image

### DIFF
--- a/core/ui/block_space/flyout.js
+++ b/core/ui/block_space/flyout.js
@@ -635,6 +635,28 @@ Blockly.Flyout.prototype.show = function(xmlList) {
     this.layoutXmlToBlocks_(xmlList, blocks, gaps, margin);
   }
 
+  // If RTL, calculate the flyout content width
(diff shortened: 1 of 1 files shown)


sureshc:
Can someone in @code-dot-org/starlabs review? And also create a copy of this Pull Request to trigger a continuous integration build?

ajpal (approved):

PR #273, opened 2021-04-02
Fix lint errors

ajpal:
I'm not 100% sure how these slipped through since `lint` is supposed to be in the pre-commit.
Fixing the errors and confirmed that the precommit is running correctly for me locally:
image

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 node_modules
+coverage
 dist
 /tests/test_dependency_map.js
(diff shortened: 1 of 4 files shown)


ajpal:
fyi @marcd123 - Can you make sure that lint is running properly for you?
Looks like some (but not all) of these errors were introduced with your recent work.

epeach (approved):
LGTM - thanks for updating

PR #279, opened 2021-07-14
[Sprite Lab] Force buttons in rectangular dropdown to a new row

ajpal:
Buttons in the sprite dropdown menu are treated the same as the sprite images (they're all just `MenuItem`s)
This could lead to some awkward looking placement depending on the number of sprites:
Before
image
image

My solution here is to add empty, unselectable menu items to fill out the row in order to force the buttons to start on a new line:
After
image
image

This was the cleanest approach I could come up with, but if there's a better way, please lmk!


epeach (approved):
Looks good! I can't think of a cleaner way to do it within the scope of our blockly dropdown.

PR #280, opened 2021-07-29
[Mini toolbox] Only re-render the event stack, not the whole blockspace

ajpal:
When we add a new sprite block to the event socket, we need to update the thumbnails on any pointer blocks in the event stack. Previously, we accomplished this by re-rendering the whole blockspace, which does work, but is a lot of extra work and was a performance issue. Instead, we should just be re-rendering the event block (which also re-renders its children)

Thumbnails still update as expected:
image


madelynkasula (approved):
really great find and fix, anjali!

epeach (commented):
Looks good to me!

PR #281, opened 2021-07-29
Only render visible blocks

ajpal:
Small performance improvement to only render visible blocks. This should not result in any user-visible changes since invisible blocks shouldn't be visible anyways.

There will be a corresponding change in the code-dot-org repo to make sure spritelab behavior definition blocks are marked as not user visible, so we don't have to render them.


madelynkasula (approved):

epeach:
Looks good! :+1:

PR #287, opened 2021-12-03
Serialize to <field> and deserialize from <field>

ajpal:
This change in the fork will facilitate migration to Google Blockly. Google Blockly renamed `<title>` to `<field>` in 2013. Currently, we have an override in our Google Blockly wrapper layer to make sure we still serialize XML using `<title>` tags [here](https://github.com/code-dot-org/code-dot-org/blob/staging-next/apps/src/blockly/addons/cdoXml.js#L30)
Instead, we should change our fork to match Google Blockly's behavior. With this change, Cdo Blockly will serialize XML using `<field>` tags, and then we can remove the override of `fieldToDom_` to that Google Blockly also serializes using `<field>` tags. Both versions of Blockly will still be able to deserialize XML that uses `<title>` tags, which is important for backwards compatibility.

Before
image

After
image
Import from XML still works with `<title>` elements
image
Import from XML also works with `<field>` elements
image

### DIFF
--- a/core/utils/xml.js
+++ b/core/utils/xml.js
@@ -69,7 +69,14 @@ Blockly.Xml.blockToDom = function(block, ignoreChildBlocks) {
   }
   function titleToDom(title) {
     if (title.name && title.EDITABLE) {
-      var container = goog.dom.createDom('title', null, title.getValue());
+      /**
+       * <title> was renamed to <field> in Google Blockly in 2013.
+       * By creating <field> elements here as well, we will make it easier to
+       * migrate to using Google Blockly everywhere.
+       * Both Google Blockly and this fork will still be able to deserialize
+       * block XML that has <title> tags.
+       */
+      var container = goog.dom.createDom('field', null, title.getValue());
       container.setAttribute('name', title.name);
       if (title.config) {
         container.setAttribute('config', title.config);
@@ -444,6 +451,9 @@ Blockly.Xml.domToBlock = function(blockSpace, xmlBlock) {
         }
         break;
       case 'title':
+      // title was renamed to field. See comment in titleToDom for more detail.
+      // fall through
+      case 'field':
         /**
          * XML example:
          * <block type="draw_move_by_constant">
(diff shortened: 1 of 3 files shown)


madelynkasula (approved):
🎉 yay for moving toward consistency! great tests as well

core/utils/xml.js
```diff
@@ -69,7 +69,14 @@ Blockly.Xml.blockToDom = function(block, ignoreChildBlocks) {
...
   function titleToDom(title) {
     if (title.name && title.EDITABLE) {
-      var container = goog.dom.createDom('title', null, title.getValue());
+      /**
+       * <title> was renamed to <field> in Google Blockly in 2013.
+       * By creating <field> elements here as well, we will make it easier to
+       * migrate to using Google Blockly everywhere.
+       * Both Google Blockly and this fork will still be able to deserialize
+       * block XML that has <title> tags.
+       */
```
katleiahramos, line 78: I think it would be helpful for the information in this JSDoc to be for the function (right above line 70). At first I was confused and thought this was supposed to be `fieldToDom` or that we should be using `field` instead of `title` but then re-read the description and realized that this is here specifically for backwards compatibility with `<title>` elements (if I'm understanding correctly).  

I think we could still leave a comment here for line 79 that mentions that we want to `use <field> element since <title> is deprecated.` or something.

tests/xml_test.js
```diff
@@ -56,3 +56,30 @@ function test_block_title_with_id() {
...
+
+  var block = new Blockly.Block(blockSpace, 'text');
+  block.initSvg();
+  block.setTitleValue('field value', 'TEXT');
+  block.getTitle_('TEXT').id = 'field id';
+  var xmlString = Blockly.Xml.domToText(Blockly.Xml.blockToDom(block));
+  var expectedXml =
+    '<block type="text"><field name="TEXT" id="field id">field value</field></block>';
+  assertEquals(xmlString, expectedXml);
+}
```
katleiahramos, line 72: :chefs-kiss: note to self,  could be useful to add to a doc as an example of set up of a Block xml test.

PR #289, opened 2022-01-03
Style updates for FieldIcon

ajpal:
refactor the style rules for hovering over the FieldIcon so that we can also make it "readonly" and the hover rules won't apply

No hover state:
image

Hover state:
image


core/ui/fields/field_icon.js
```diff
@@ -25,3 +25,8 @@ Blockly.FieldIcon.prototype.EDITABLE = false;
  * @private
  */
 Blockly.FieldIcon.prototype.showEditor_ = function() {};
+
+Blockly.FieldIcon.prototype.setReadOnly = function() {
+  Blockly.addClass_(this.fieldGroup_, 'readonly');
+  this.textElement_.setAttribute('style', 'font-size:9pt; cursor:default;');
```
madelynkasula, line 31: should/could these styles live in a file like core/ui/css.js?

madelynkasula (approved):

PR #311, opened 2023-02-24
[Student Learning] [SL-207] Fix Custom behavior dialog's green background being misplaced

levadadenys:
### [Student Learning] [SL-20 7] Fix Custom behavior dialog's green background being misplaced

Before:
Знімок екрана 2023-02-24 о 19 20 59
Знімок екрана 2023-02-24 о 19 17 29

After:
Знімок екрана 2023-02-24 о 19 16 34
Знімок екрана 2023-02-24 о 19 17 47
image

- jira ticket: [SL-207](https://codedotorg.atlassian.net/browse/SL-207?atlOrigin=eyJpIjoiMTY3NWQ0ZDI5OTllNDVhYmI5NjAxMzI3MzJhMzFlZDUiLCJwIjoiaiJ9)

### DIFF
--- a/core/ui/function_editor.js
+++ b/core/ui/function_editor.js
@@ -1039,6 +1039,7 @@ Blockly.FunctionEditor.prototype.setupParametersToolbox_ = function() {
 };
 
 Blockly.FunctionEditor.prototype.addEditorFrame_ = function() {
+  var noToolboxPadMagicNumber = 36;
   // if we are in readOnly mode, don't pad. Otherwise, pad left
   // based on the size of either the Toolbox or the Flyout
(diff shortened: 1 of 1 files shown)


mikeharv (approved):
It's unfortunate that we haven't found a root cause, but considering this is broken on production, I think it makes sense to merge this. 

It's probably worth noting that because we're working to move labs off of this version of Blockly that I wouldn't expect us to make any other fixes here later on, unless of course something else was found to be broken.

mikeharv:
@levadadenys Since this hasn't merged yet, would you mind just confirming that this works as expected for these levels in RTL orientation?

PR #315, opened 2023-03-07
Simplify logic for function editor frame

mikeharv:
Addresses this eyes diff:
Screenshot 2023-03-07 at 11 16 17 AM

In the above case, we couldn't use `getToolboxWidth()`, because it turns out that it's calculated while the Variables flyout is still open. However, the new logic we were using to set the padding for levels without a toolbox actually works just fine even when we do have one!

Fix history:
* https://github.com/code-dot-org/blockly/pull/311
* https://github.com/code-dot-org/blockly/pull/314

The most recent attempt to fix the function editor frame introduced yet another new regression - this time in CSinAlg, which I didn't think to test manually. As a result, I reverted a Blockly bump in our cdo repo. Once this fix is merged, I will publish a new patch version and bump to that instead.

### DIFF
--- a/core/ui/function_editor.js
+++ b/core/ui/function_editor.js
@@ -1039,19 +1039,16 @@ Blockly.FunctionEditor.prototype.setupParametersToolbox_ = function() {
 };
 
 Blockly.FunctionEditor.prototype.addEditorFrame_ = function() {
-  var noToolboxPad =
+  var toolboxPad =
     Blockly.modalBlockSpace.getMetrics().absoluteLeft -
     FRAME_MARGIN_SIDE -
     Blockly.Bubble.BORDER_WIDTH -
     1;
   // if we are in readOnly mode, don't pad. Otherwise, pad left
-  // based on the size of either the Toolbox or the Flyout
-  var toolboxWidth = Blockly.mainBlockSpace.blockSpaceEditor.getToolboxWidth();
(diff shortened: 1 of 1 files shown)


core/ui/function_editor.js
```diff
@@ -1039,19 +1039,16 @@ Blockly.FunctionEditor.prototype.setupParametersToolbox_ = function() {
...
-  // based on the size of either the Toolbox or the Flyout
-  var toolboxWidth = Blockly.mainBlockSpace.blockSpaceEditor.getToolboxWidth();
-  var left = this.modalBlockSpace.isReadOnly()
-    ? 0
-    : toolboxWidth
-    ? toolboxWidth
-    : noToolboxPad;
+  // based on the logic above.
+  // Previously we just used the size of either the Toolbox or the Flyout
+  // but this was causing problems in levels with blocks with their own flyouts.
```
breville, line 1050: out of curiosity: a block can have its own flyout?

breville (approved):

core/ui/function_editor.js
mikeharv: Yes! It was one of the cooler features that Anjali built for Sprite Lab (and one we'll eventually need to redo to work with mainline). There are a few event handling blocks that let you point to the specific sprite that was involved in an event - rather than selecting all of the sprites that are implicitly part of the same costume-group: 
minitoolbox

PR #316, opened 2023-03-13
add support for behaviorPicker blocks

mikeharv:
Adds support to the function editor for new `behaviorPicker` blocks in Sprite Lab. This will only be merged if the following is approved:
- https://github.com/code-dot-org/code-dot-org/pull/50702

Once this is merged, a new release will need to be created. A bump in the code-dot-org repo should accompany the above PR.

### DIFF
--- a/core/ui/function_editor.js
+++ b/core/ui/function_editor.js
@@ -20,6 +20,56 @@ goog.require('goog.structs.LinkedMap');
 /** @const */ var FRAME_MARGIN_TOP = 10;
 /** @const */ var FRAME_HEADER_HEIGHT = 25;
 
+function getAllBehaviorPickerBlocks() {
+  var allBlocks = Blockly.mainBlockSpace.getAllBlocks();
+  var allBehaviorPickerBlocks = allBlocks.filter(function(block) {
+    return block.type === 'gamelab_behaviorPicker';
+  });
+  return allBehaviorPickerBlocks;
+}
+
+function updateBehaviorPickerFields(block, behaviorId, newValue) {
+  if (block.type !== 'gamelab_behaviorPicker') {
+    return;
+  }
+  var fieldValue = block.getFieldValue('BEHAVIOR');
+  if (
+    getAllBehaviorsIds().indexOf(fieldValue) === -1 &&
+    getAllBehaviorsIds().indexOf(newValue) > -1
+  ) {
+    // Update user-defined behaviors based on the updated value
+    block.setTitleValue(newValue, 'BEHAVIOR');
+  } else if (
+    fieldValue === behaviorId &&
+    getAllBehaviorsIds().indexOf(behaviorId) > -1
+  ) {
+    // Refresh non-user-defined behaviors based on the persistent id value
+    block.setTitleValue(behaviorId, 'BEHAVIOR');
+  }
+}
+
+function resetDeletedBehaviorPickerFields(block, behaviorName) {
+  if (block.type !== 'gamelab_behaviorPicker') {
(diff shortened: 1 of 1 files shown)


core/ui/function_editor.js
```diff
@@ -20,6 +20,56 @@ goog.require('goog.structs.LinkedMap');
...
+  });
+  return allBehaviorPickerBlocks;
+}
+
+function updateBehaviorPickerFields(block, behaviorId, newValue) {
+  if (block.type !== 'gamelab_behaviorPicker') {
+    return;
+  }
+  var fieldValue = block.getFieldValue('BEHAVIOR');
+  if (
```
molly-moen, line 36: is there ever a case where neither the if or the else if would evaluate to true? And if so, is our expected behavior to do nothing?

core/ui/function_editor.js
```diff
@@ -20,6 +20,56 @@ goog.require('goog.structs.LinkedMap');
...
 
+function getAllBehaviorPickerBlocks() {
+  var allBlocks = Blockly.mainBlockSpace.getAllBlocks();
+  var allBehaviorPickerBlocks = allBlocks.filter(function(block) {
+    return block.type === 'gamelab_behaviorPicker';
+  });
+  return allBehaviorPickerBlocks;
+}
+
+function updateBehaviorPickerFields(block, behaviorId, newValue) {
```
molly-moen, line 31: can you add a comment to this method and `resetDeletedBehaviorPickerFields` explaining what they do?

core/ui/function_editor.js
mikeharv: Yes, that would be relatively common.
Consider the following student code:
image


If "new behavior" (user-defined behavior) is changed, the first picker block should get updated, but not the other two - we detect that the old value isn't a behavior but the new one one is.  If "moving east" (shared behavior with a constant id) is changed, then both the blocks that list that behavior will update - the field value matches the current behavior id. 

In the first case, we need to update the selected option to use the new option (array of user-facing string and actual id). For the second case, we already have the right id, but selecting it again will update the field with the correct/updated user-facing string. The user-facing string is the behavior "name", while the id correlates to the actual generated code.

core/ui/function_editor.js
mikeharv: Done. For this method, I also re-ordered the clauses to align to the parameter order and their use. I also added returns and en empty `else` clause with a commented return to better convey the intentionality of the structure.

core/ui/function_editor.js
```diff
@@ -20,6 +20,68 @@ goog.require('goog.structs.LinkedMap');
...
+  if (
+    fieldValue === behaviorId &&
+    getAllBehaviorsIds().indexOf(behaviorId) > -1
+  ) {
+    // Refresh non-user-defined behaviors based on the persistent id value.
+    // Selecting the same id again will update the field with the correct user-facing name.
+    block.setTitleValue(behaviorId, 'BEHAVIOR');
+    return;
+  } else if (
+    getAllBehaviorsIds().indexOf(fieldValue) === -1 &&
```
molly-moen, line 48: for my understanding: is the index of the field value -1 because it was a user-defined behavior that is no longer in the list?

molly-moen (approved):
thanks for the additional explanations!

core/ui/function_editor.js
mikeharv: That's correct. I had wanted to just use `.includes()` here, but this file only supports ES5 and that was the best workaround I could find.

PR #317, opened 2023-03-22
[DES-260] Remove .red-delete-button

levadadenys:
### [DES-260] Remove .red-delete-button
Jira [DES-260](https://codedotorg.atlassian.net/browse/DES-260?atlOrigin=eyJpIjoiYzNiOWMyY2ZlYTE1NDg2OWJmYWFmNmU3OTU2NjAwZTgiLCJwIjoiaiJ9)

Part of https://github.com/code-dot-org/code-dot-org/pull/50918

The main difficulty here is that we've had some wrong styles and parameters set in Blockly, however they won't work as expected without the changes made in CDO PR cause that PR along with bumping Blockly version creates a fix for that edge case that we initially had to update in Blockly.
 
And some more info for better understanding: 
The main problem here is that we already have a set of dialogs and button defined in blockly. While doing Phase 1.5 I was simply updating the cdo styles expecting all the dialogues to be consistent. But it turned out that they are not as consistent and not always used correct naming/structure/styles, just like the button I'm working on in this pr. It was a cancel button that was painted in red with the help of injected css with !important.

So, in short, the idea here is to fix and simplify where possible what we have right now without going too deep and rewriting too much of cdo-blockly (keeping in mind that we're going to get rid of it at some place)
 
Before:
image

After:

**(without CDO PR)**
image

**(WITH CDO PR)**
image


mikeharv:
The code changes make sense here, but I'm surprised to learn that these buttons are supposed to be white and not red. Originally, we used red for buttons like this to indicate that users should be careful (because something will be deleted). What's our updated thinking on buttons with this kind of purpose? Are they referenced in any of the design specs/docs?

levadadenys:
> The code changes make sense here, but I'm surprised to learn that these buttons are supposed to be white and not red. Originally, we used red for buttons like this to indicate that users should be careful (because something will be deleted). What's our updated thinking on buttons with this kind of purpose? Are they referenced in any of the design specs/docs?

yeah, I was thinking about that. We can make it danger button instead of cancel button then. It'll look like this:
image

mikeharv:
This change looks good provided that we only bump to the new version along with https://github.com/code-dot-org/code-dot-org/pull/50918

Thanks for all the back and forth communication and for answering all of my questions offline!

mikeharv (approved):

levadadenys:
I'm so sorry. I've forgot to push 1 extra commit. Reverting this PR and creating new PR we this and 1 extra commit. The description of it will be the same and still actuall.

PR #323, opened 2023-11-14
CT-175: Stop using function ids for function definitions

ebeastlake:
Jira tickets:
- https://codedotorg.atlassian.net/browse/P20-522
- https://codedotorg.atlassian.net/browse/CT-175

Platform discovered a translation bug due to an inconsistency where function definition blocks are assigned an ID, whereas function call blocks are not. My understanding is that when a function is defined in a translated language, the corresponding function call block cannot associate with the definition block due to the ID inconsistency, causing the program to crash or freeze.

The ID is added to function definition blocks when the modal function editor is enabled but not to the corresponding function call blocks. The ID addition is necessary for shared behaviors but is unintentionally applying to function blocks as well.

@mike and @molly-moen identified two potential solutions:
1. Translate function ids
2. Ignore function ids when we are generating code (always use the name)

For now, we decided to not use IDs for functions, because translating them adds extra complexity for a case we don’t really want to support.

### DIFF
--- a/core/ui/function_editor.js
+++ b/core/ui/function_editor.js
@@ -782,7 +782,8 @@ Blockly.FunctionEditor.prototype.create_ = function() {
       this.container_.querySelector('#functionNameText').value = value;
     }
     this.functionDefinitionBlock.setTitleValue(value, 'NAME');
-    if (this.functionDefinitionBlock.type === 'behavior_definition') {
+    console.log("this.functionDefinitionBlock.type", this.functionDefinitionBlock.type);
+    if (this.functionDefinitionBlock.userCreated && this.functionDefinitionBlock.type === 'behavior_definition') {
       this.functionDefinitionBlock.getTitle_('NAME').id = value;
     }
     if (Blockly.Blocks.gamelab_behaviorPicker) {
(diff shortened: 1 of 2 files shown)


core/ui/function_editor.js
```diff
@@ -782,9 +782,6 @@ Blockly.FunctionEditor.prototype.create_ = function() {
       this.container_.querySelector('#functionNameText').value = value;
     }
     this.functionDefinitionBlock.setTitleValue(value, 'NAME');
-    if (this.functionDefinitionBlock.userCreated) {
```
molly-moen, line 785: I think we might still need this if this is a behavior? See [this PR](https://github.com/code-dot-org/blockly/pull/257)

mikeharv: +1 I think we can just further check that the type of functionDefintionBlock so it only happens for behaviors.

ebeastlake: Ah, thanks! For some reason, I thought the behavior piece of this was handled elsewhere.

mikeharv: To clarify, we also need to keep the userCreated check. Otherwise, we’ll start changing the ids of shared behaviors when they are renamed or translated. We only want to set the id here if it’s the student’s own behavior.

molly-moen (approved):
Nice!

mikeharv (approved):

